PR #121, opened 2017-11-29
Change instance count from 3 to 1

gurumurthys:


nadiaya:
Why is this change necessary?

gurumurthys:
Many customers have a limit on the number of p2.x machines (usually 1 or 2) and hence it may fail for them.

PR #128, opened 2017-12-01
updates sagemaker-runtime reference

mikevansnell:
looks like this library name changed. This allows the breast cancer predictor to work properly.


djarpin:
Thanks, Mike for your interest in Amazon SageMaker Examples!

There is currently a slight misalignment between the "sample-notebooks" that are available in an Amazon SageMaker Notebook Instance and these examples on GitHub.  In this case, the GitHub repo is actually ahead of the Notebook Instances and is in alignment with the external AWS CLI and boto3 library.  A change is being made in the next day or two so that the "sample-notebooks" that appear in a Notebook Instance align and use "runtime.sagemaker" as well.

PR #129, opened 2017-12-01
TF IRIS BYOM is ready.

ragavvenkatesan:


under_development/tensorflow_iris_byom/tensorflow_BYOM_iris.ipynb
```diff
@@ -68,47 +68,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "import os\n",
```
djarpin: Typos in markdown above... "mdoe" instead of "mode".  And "enstimator" should be "estimator".  Also instead of "Supply the role in the variable below", let's go with "If the get_execution_role() function does not return a role with the appropriate permissions, you'll need to specify an IAM role arn that does."

under_development/tensorflow_iris_byom/tensorflow_BYOM_iris.ipynb
```diff
@@ -341,34 +310,32 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {
```
djarpin: In the above:
- Replace "instance for this training" with "instance for this example."  Just to avoid confusion with the model training process.
- Can we consistently expect 1?  I got 2 with a score of 0.5236916542053223.  Didn't have time to dig in but I expect it's due to initial randomization.
- Instead of "Run a sample prediction on a sample..." can we change to "Run and example prediction on a sample..." just to avoid confusion?
- Change "continied use of" to "to continue using".

under_development/tensorflow_iris_byom/tensorflow_BYOM_iris.ipynb
```diff
@@ -220,8 +182,15 @@
    "metadata": {},
```
djarpin: In markdown above:
- Let's replace "export\Servo" backslashes with forward slashes "export/Servo"

ragavvenkatesan:
All suggestions incorporated.

ragavvenkatesan:
I am not sure how to move this directly into the advanced_functionality directory. I think this might affect some path issues with readme as well. I have an older pull of this repo in which I don't have these directories.

PR #134, opened 2017-12-05
Make the MXNet Gluon MNIST Example scale

aidan-plenert-macdonald:
Added optional distributed KV store

### DIFF
--- a/sagemaker-python-sdk/mxnet_gluon_mnist/mnist.py
+++ b/sagemaker-python-sdk/mxnet_gluon_mnist/mnist.py
@@ -2,7 +2,7 @@ from __future__ import print_function
 
 import logging
 import mxnet as mx
-from mxnet import gluon, autograd
+from mxnet import gluon, autograd, kv
 from mxnet.gluon import nn
 import numpy as np
 import json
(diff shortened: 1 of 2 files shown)


sagemaker-python-sdk/mxnet_gluon_mnist/mnist.py
```diff
@@ -41,8 +41,10 @@ def train(channel_input_dirs, hyperparameters, **kwargs):
     # Collect all parameters from net and its children, then initialize them.
     net.initialize(mx.init.Xavier(magnitude=2.24), ctx=ctx)
     # Trainer is for updating parameters with gradient.
+    store = kv.create('dist_sync' if len(hosts) > 1 else 'local')
```
jesterhazy: thanks for this PR!

instead of importing kv, it would be simpler to just use the `str` version:

`kvstore = 'dist_sync' if len(hosts) > 1 else 'local'`

also, if you use `dist_device_sync` and `device` instead, then this will use device mode when gpus are available, and fall back to cpu (dist_sync or local) when they aren't.

aidan-plenert-macdonald: Done. Also fixed in `mxnet_gluon_sentiment/sentiment.py`

PR #135, opened 2017-12-05
Image classification - added explanation for incorrect output

gurumurthys:
Added description of output and why it is incorrect. Suggested to train for more epochs to get better output

### DIFF
--- a/introduction_to_amazon_algorithms/imageclassification_caltech/Image-classification-fulltraining.ipynb
+++ b/introduction_to_amazon_algorithms/imageclassification_caltech/Image-classification-fulltraining.ipynb
@@ -532,7 +532,7 @@
     "\n",
     "Evaluate the image through the network for inteference. The network outputs class probabilities and typically, one selects the class with the maximum probability as the final class output.\n",
     "\n",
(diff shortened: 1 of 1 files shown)


introduction_to_amazon_algorithms/imageclassification_caltech/Image-classification-fulltraining.ipynb
```diff
@@ -524,6 +524,17 @@
...
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Evaluation\n",
+    "\n",
+    "Evaluate the image through the network for inteference. The network outputs class probabilities and typically, one selects the class with the maximum probability as the final class output.\n",
+    "\n",
+    "**Note:** The output class detected by the network may not be accurate in this example. This is because we have trained the model only for a couple of epochs. If the network is trained for more epochs (say 20), then the output class will be more accurate."
```
djarpin: Can we change "This is because we have..." to something like "To limit the time and cost of training, we have...".  Just want to make sure the customer understands why we've chosen to only train for 2 epochs.

PR #152, opened 2017-12-19
Fixed: build_and_push.sh to include docker restart command

djarpin:


tomfaulhaber:
This will kill any Docker containers that the user has running

PR #158, opened 2018-01-08
Changed: Kernel for Breast Cancer Prediction.ipynb

djarpin:


lynnlangit:
For context, I am an AWS Community Hero and also I've been doing work in this area with CSIRO Bioinformatics - [article](https://databricks.com/blog/2017/07/26/breaking-the-curse-of-dimensionality-in-genomics-using-wide-random-forests.html).

My feedback for your example notebook is that if it would be possible to add a couple of visualizations the example would be more broadly understandable (and by association - usable).  

HTH 

--Lynn

djarpin:
@lynnlangit 
Thanks for your feedback and interest in our example notebooks!  We're always looking for ways to make them better, so this helps.  We'll keep this pull request small, just to make the kernel change, but will look to add more visualizations to the notebook going forward.  Also, if you have any very specific ideas in mind, we love getting contributions from the community, so feel free to open up a pull request!

PR #159, opened 2018-01-08
Update xgboost_direct_marketing_sagemaker.ipynb

roachmd:
I was getting an error with this line in conda_python3 :: hist = data.hist(bins=30, sharey=True, figsize=(10, 10))
I updated it to read hist = data.hist(30, True, figsize(10, 10))


roachmd:
Running this in a new sagemaker notebook.
pd.__version__
0.20.3

That line I get "SyntaxError: positional argument follow keyword argument"

Now I just remove and retyped the line. Seems fine now. So, disreguard ;-)

PR #185, opened 2018-02-02
Fixed: XGBoost direct marketing.  Train with CSV and update markdown.

djarpin:
Tested in ml.m4.xlarge instance and runs correctly.

### DIFF
--- a/introduction_to_applying_machine_learning/xgboost_direct_marketing/xgboost_direct_marketing_sagemaker.ipynb
+++ b/introduction_to_applying_machine_learning/xgboost_direct_marketing/xgboost_direct_marketing_sagemaker.ipynb
@@ -511,7 +511,9 @@
     "## Evaluation\n",
     "There are many ways to compare the performance of a machine learning model, but let's start by simply comparing actual to predicted values.  In this case, we're simply predicting whether the customer subscribed to a term deposit (`1`) or not (`0`), which produces a simple confusion matrix.\n",
     "\n",
-    "First we'll need to determine how we pass data into and receive data from our endpoint.  Our data is currently stored as NumPy arrays in memory of our notebook instance.  To send it in an HTTP POST request, we'll serialize it as a CSV string and then decode the resulting CSV."
+    "First we'll need to determine how we pass data into and receive data from our endpoint.  Our data is currently stored as NumPy arrays in memory of our notebook instance.  To send it in an HTTP POST request, we'll serialize it as a CSV string and then decode the resulting CSV.\n",
+    "\n",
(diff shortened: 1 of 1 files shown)


mattflo-outpace:
so, csv is working fine, but there's a problem with libsvm?

mattflo-outpace:
yep. csv seems to work fine!

introduction_to_applying_machine_learning/xgboost_direct_marketing/xgboost_direct_marketing_sagemaker.ipynb
```diff
@@ -348,18 +356,19 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Amazon SageMaker's XGBoost container expects data in the libSVM data format.  This expects features and the target variable to be provided as separate arguments.  Let's split these apart.  Notice that although repetitive it's easiest to do this after the train|validation|test split rather than before.  This avoids any misalignment issues due to random reordering."
+    "Amazon SageMaker's XGBoost container expects data in the libSVM or CSV data format.  For this example, we'll stick to CSV.  Note that the first column must be the target variable and the CSV should not include headers.  Also, notice that although repetitive it's easiest to do this after the train|validation|test split rather than before.  This avoids any misalignment issues due to random reordering."
```
ssaini4: For training, SageMaker XGBoost expects the CSV and LibSVM datasets to have the target variable in the first column. 
For testing, SageMaker XGBoost expects CSV datasets to NOT have target variables. Only features are required. However, for LibSVM the test dataset may or may not have the target variables in the first column

djarpin: Updated the markdown, but did so closer to the inference section where I think it's more relevant.

mattflo-outpace: So, was this a change? The livsvm version of this example notebook clearly removed the target column? If this was a change, is there somewhere we should be looking for such important changes?

djarpin: XGBoost always supported LibSVM and CSV as training and inference formats.  And CSV inference always required the target column to be removed.  The wording change is just there to provide additional detail and more clarity on the topic.  We're still investigating what drove the problem in the earlier version of the notebook.  So, no changes to the XGBoost algorithm yet.  Thanks for your interest.

PR #186, opened 2018-02-02
Updates: BYO examples to remove docker restart

djarpin:
Tested both notebooks and build_and_push.sh in fresh SageMaker Notebook Instance and both worked.


advanced_functionality/scikit_bring_your_own/container/build_and_push.sh
```diff
@@ -13,6 +13,9 @@ then
     exit 1
 fi
 
+chmod +x decision_trees/train
+chmod +x decision_trees/serve
+
```
mckev-amazon: Is this a related change?

djarpin: I snuck that one in there.  It was a point of misalignment between the notebook's shell command and the build_and_push.sh that was called out in this [Issue](https://github.com/awslabs/amazon-sagemaker-examples/issues/179).

mckev-amazon: Sounds good!  Not an issue, from my perspective, as long as both issues are referenced by the commit (which you've done).

PR #224, opened 2018-03-31
Fixed: payload was larger than SageMaker limit

djarpin:


djarpin:
Yes.  I tested 100 and it worked.  I figured going to an even slightly smaller payload (100 splits rather than 60 would give us some headroom/future-proofing).

PR #239, opened 2018-04-19
 Update the helper function to cat is not Non

qx0731:
Otherwise, cat = 0 won't be written to the data file


qx0731:
check ther cat to allow cat = 0

PR #272, opened 2018-06-07
Add deepar_retail_forecasting example

robperc:
This example shows a more compelling real world use-case of DeepAR where the algorithm is used to predict clothing retail product sales using a synthetic retail dataset.


robperc:
Fixed some bugs introduced due to an update to the pandas version used by the notebook.

eoghanmull:
I have a quick query on the model output i.e. "mean" vs "quantiles". In this example you use the quantile of 0.9, in other examples I have seen the use of 0.5 and also the separate output parameter "mean". Which one should be used as the "prediction" is there a rule of thumb or is it a more case by case basis.

robperc:
The "mean" output type provides a single a single output series that is the mean of the number of samples specified in the "num_samples" configuration parameter. The "quantiles" output type returns all series for the samples that fall into the quantiles specified in the "quantiles" configuration parameter.

eoghanmull:
Thank you Rob, I think I may have phrased the question incorrectly, what I'm trying to figure out is what the quantiles signify.

My current intuition is that each quantile is a bias for the predictions trend i.e 0.9 signifies an "over prediction" and 0.1 an "under prediction" is this correct?

Kerstin123:
Thank you for your example:)

I have a question regarding the cat-operator you integrated in your code.
You have two categories but only one cat in your final json file. Is this category a combination you generated in your code which contains a combination of both categories?
And do you know a way to imlement the categories seperately?

I tried to implement the categories seperatly within your code but I got an error message that the training set did not contain examples for all categories of the 1th categorical feature.

I think the reason is that the encodeing of the categories is a loop over all categories but it should be seperate for each one.

Do you know a way to encode all categories serperately?

Thanks in advance

hsl89:
Dear SageMaker Community,
             In our attempts to provide this repository with a better level of support             going forward, we’re closing issues that were opened prior to the v2 release of the SDK.             This is because we believe that over time many of the issues             posted were solved with the latest release or other recent changes to the repo.             This will help us reallocate resources towards issues that are            more likely to still be relevant today.            Some of the issues experienced now can be resolved by referencing the            v2 guide: https://sagemaker.readthedocs.io/en/stable/v2.html             In this guide, you can find simple solutions to common notebook errors,             like the renaming of parameters and classes.            If you believe your issue is still ongoing and you have updated error            messaging or other info, please re-open it, and we investigate the issue.
            Best Regards,
            AWS SageMaker Team

PR #282, opened 2018-06-12
change output data dir for chainer

andremoeller:
The value of SM_OUTPUT_DATA_DIR is changing in the container. This needs to be deployed before the Chainer container is deployed or these notebooks will break.


andremoeller:
@djarpin ,

I just pushed the MNIST changes, thanks!

PR #297, opened 2018-06-29
Adding a framework version so it doesn't fail.

awan-10:
This shall fix the errors of the form:
```
ValueError: Error training tensorboard-example-2018-06-28-23-02-44-061: Failed Reason: AlgorithmError: uncaught exception during training: 'module' object has no attribute 'FixedLengthRecordDataset
```


nadiaya:
Thank you!

PR #330, opened 2018-07-17
minor change to object detection notebook

EvanzzzZ:
Minor change to object detection notebook:
- Use key 'prediction' to extract values from responses


EvanzzzZ:
@djarpin Can you merge this change

PR #333, opened 2018-07-19
object_detection fix visualization

ranman:
This commit makes the visualization this notebook work correctly.

Currently the endpoint returns an object that is wrapped in a `{'prediction': []}` object. The visualization does expects the raw predictions array.


djarpin:
Thanks, @ranman - Sorry it took me so long to get to this one.  Closing since this should be covered by the recently merged PR #330 .

PR #339, opened 2018-07-20
imports

samsammurphy:
'boto3' imported twice, 're' module import could probably be in the main import block


advanced_functionality/kmeans_bring_your_own_model/kmeans_bring_your_own_model.ipynb
```diff
@@ -77,9 +76,9 @@
     "import urllib.request\n",
     "import json\n",
     "import mxnet as mx\n",
-    "import boto3\n",
```
djarpin: Any particular concern with importing boto3 twice?  The advantage of having it in both cells is that if all your imports are in one cell, but you don't use the get_execution_role() function, you don't have to run it for future boto3 calls to work.

PR #351, opened 2018-07-31
Added short note about validation channel

pigna90:
According to [this issue](https://github.com/awslabs/amazon-sagemaker-examples/issues/334) I added a simple note to explain why the validation channel is not explicitly used.

### DIFF
--- a/introduction_to_amazon_algorithms/xgboost_abalone/xgboost_abalone.ipynb
+++ b/introduction_to_amazon_algorithms/xgboost_abalone/xgboost_abalone.ipynb
@@ -284,6 +284,13 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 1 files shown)


introduction_to_amazon_algorithms/xgboost_abalone/xgboost_abalone.ipynb
```diff
@@ -559,9 +566,9 @@
  "metadata": {
   "anaconda-cloud": {},
   "kernelspec": {
-   "display_name": "conda_python3",
+   "display_name": "Python 3",
    "language": "python",
-   "name": "conda_python3"
+   "name": "python3"
```
djarpin: It looks like you may have made this edit in an environment outside SageMaker notebooks.  Unfortunately, because SageMaker notebook instances don't have a "python3" kernel (instead it's called "conda_python3"), this notebook will open with an error and the user will have to select a different kernel.  We'd like to avoid that.  Can you resubmit with the kernelspec updated to its original values?  Thanks.

PR #357, opened 2018-08-06
Very minor spelling correction

nedlowe:


nedlowe:
Ah I see someone else has made the same correction and month ago, I'll withdraw this.

PR #365, opened 2018-08-08
Extending SageMaker PyTorch containers

ChoiByungWook:
Example showing how to use our PyTorch container as a base image.


advanced_functionality/pytorch_extending_our_containers/pytorch_extending_our_containers.ipynb
```diff
@@ -0,0 +1,701 @@
...
+    "Even if there is direct SDK support for your environment or framework, you may want to add additional functionality or configure your container environment differently while utilizing our container to use on SageMaker.\n",
+    "\n",
+    "All of our deep learning framework containers are open source on GitHub. Based on your use case it you can easily modify the container code and build a modified version of our containers.\n",
+    "\n",
+    "* [SageMaker TensorFlow Container](https://github.com/aws/sagemaker-tensorflow-container)\n",
+    "* [SageMaker MXNet Container](https://github.com/aws/sagemaker-mxnet-container)\n",
+    "* [SageMaker PyTorch Container](https://github.com/aws/sagemaker-pytorch-container) \n",
+    "* [SageMaker Chainer Container](https://github.com/aws/sagemaker-chainer-container)\n",
+    "\n",
+    "This walkthrough shows that it is quite straightforward to extend one of our containers to build your own custom container.\n",
```
mvsusp: what about we mention the byoc tf example here and compare the differences?

advanced_functionality/pytorch_extending_our_containers/container/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+
+# For more information on creating a Dockerfile
+# https://docs.docker.com/compose/gettingstarted/#step-2-create-a-dockerfile
+# SageMaker PyTorch image
+FROM 520713654638.dkr.ecr.us-west-2.amazonaws.com/sagemaker-pytorch:0.4.0-cpu-py3
```
mvsusp: What about add a link to https://github.com/awslabs/amazon-sagemaker-examples/master/advanced_functionality/pytorch_extending_our_containers/pytorch_extending_our_containers.ipynb

advanced_functionality/pytorch_extending_our_containers/pytorch_extending_our_containers.ipynb
```diff
@@ -0,0 +1,701 @@
...
+    "6. python_version - The supported version of Python.\n",
+    "\n",
+    "So the SageMaker PyTorch ECR image would be:\n",
+    "520713654638.dkr.ecr.us-west-2.amazonaws.com/sagemaker-pytorch:0.4.0-cpu-py3\n",
+    "\n",
+    "Information on supported frameworks and versions can be found in this [README](https://github.com/aws/sagemaker-python-sdk).\n",
+    "\n",
+    "Next, we add the code that implements our specific algorithm to the container and set up the right environment for it to run under.\n",
+    "\n",
+    "**DISCLAIMER: As of now, the support for the two environment variables below are only supported for the SageMaker Chainer (4.1.0+) and PyTorch (0.4.0+) containers.**\n",
```
mvsusp: The variables would work if you add RUN pip install sagemaker-containers, right?

advanced_functionality/pytorch_extending_our_containers/container/cifar10/cifar10.py
```diff
@@ -0,0 +1,171 @@
...
+    parser.add_argument('--epochs', type=int, default=2, metavar='E',
+                        help='number of total epochs to run (default: 2)')
+    parser.add_argument('--batch-size', type=int, default=4, metavar='BS',
+                        help='batch size (default: 4)')
+    parser.add_argument('--lr', type=float, default=0.001, metavar='LR',
+                        help='initial learning rate (default: 0.001)')
+    parser.add_argument('--momentum', type=float, default=0.9, metavar='M', help='momentum (default: 0.9)')
+    parser.add_argument('--dist-backend', type=str, default='gloo', help='distributed backend (default: gloo)')
+
+    env = sagemaker_containers.training_env()
```
mvsusp: You are using the training env here but you are not explaining how it works anywhere. Consider add a little burb about this method in the notebook and link to this https://github.com/aws/sagemaker-containers/blob/master/src/sagemaker_containers/_env.py#L385


Another option, is to use env vars and instead and link to this docs here https://github.com/aws/sagemaker-containers#how-a-script-is-executed-inside-the-container

I believe that Leo likes more the second option.

advanced_functionality/pytorch_extending_our_containers/pytorch_extending_our_containers.ipynb
ChoiByungWook: For our framework containers, not necessarily because we changed the name of methods here.
https://github.com/aws/sagemaker-containers/blob/master/src/sagemaker_containers/_modules.py#L274
https://github.com/aws/sagemaker-pytorch-container/blob/master/src/sagemaker_pytorch_container/training.py#L50

advanced_functionality/pytorch_extending_our_containers/container/cifar10/cifar10.py
ChoiByungWook: Swapped to env vars.

advanced_functionality/pytorch_extending_our_containers/pytorch_extending_our_containers.ipynb
```diff
@@ -0,0 +1,701 @@
...
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Part 1: Packaging and Uploading your Algorithm for use with Amazon SageMaker\n",
+    "\n",
+    "### An overview of Docker\n",
+    "\n",
+    "If you're familiar with Docker already, you can skip ahead to the next section.\n",
+    "\n",
+    "For many data scientists, Docker containers are a new technology. But they are not difficult and can significantly simply the deployment of your software packages. \n",
```
djarpin: s/simply/simplify/

advanced_functionality/pytorch_extending_our_containers/pytorch_extending_our_containers.ipynb
```diff
@@ -0,0 +1,701 @@
...
+    "  1. [Set up the environment](#Set-up-the-environment)\n",
+    "  1. [Create the session](#Create-the-session)\n",
+    "  1. [Upload the data for training](#Upload-the-data-for-training)\n",
+    "  1. [Training On SageMaker](#Training-on-SageMaker)\n",
+    "  1. [Optional cleanup](#Optional-cleanup)  \n",
+    "1. [Reference](#Reference)\n",
+    "\n",
+    "_or_ I'm impatient, just [let me see the code](#The-Dockerfile)!\n",
+    "\n",
+    "## When should I extend a SageMaker container?\n",
```
djarpin: Similar to the "When should I build my own" section of the [TensorFlow notebook](https://github.com/awslabs/amazon-sagemaker-examples/blob/master/advanced_functionality/tensorflow_bring_your_own/tensorflow_bring_your_own.ipynb) can we add some specific, concrete examples for what type of functionality users might want to add or what configurations they might want to change?

advanced_functionality/pytorch_extending_our_containers/pytorch_extending_our_containers.ipynb
djarpin: I think we should either add some more markdown about a workaround if users wanted to do this with the TensorFlow and MXNet containers, or I think we should remove the mention of those frameworks at the top of the notebook and make it clear that this is functionality that only applies to PyTorch and Chainer right away.

advanced_functionality/pytorch_extending_our_containers/pytorch_extending_our_containers.ipynb
```diff
@@ -0,0 +1,701 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## SageMaker Python SDK Local Training\n",
+    "To represent our training, we use the Estimator class, which needs to be configured in five steps. \n",
+    "1. IAM role - our AWS execution role\n",
+    "2. train_instance_count - number of instances to use for training.\n",
+    "3. train_instance_type - type of instance to use for training. For training locally, we specify `local`.\n",
```
djarpin: I'd mention 'local_gpu' here as well.

advanced_functionality/pytorch_extending_our_containers/pytorch_extending_our_containers.ipynb
```diff
@@ -0,0 +1,701 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from sagemaker.estimator import Estimator\n",
+    "\n",
+    "hyperparameters = {'epochs': 1}\n",
+    "\n",
+    "instance_type = 'local'\n",
```
djarpin: Since this is CIFAR, I think we should have logic to use the GPU if one is available.  e.g.

```python
instance_type = 'local'

if subprocess.call('nvidia-smi') == 0:
    ## Set type to GPU if one is present
    instance_type = 'local_gpu'
```

advanced_functionality/pytorch_extending_our_containers/pytorch_extending_our_containers.ipynb
ChoiByungWook: Yes, I added it in but it... isn't bolded or in its own section.

'Even if there is direct SDK support for your environment or framework, you may want to add additional functionality or configure your container environment differently while utilizing our container to use on SageMaker.'

There really is only one reason to extend our containers, and that is to add additional libraries or modify configurations.

advanced_functionality/pytorch_extending_our_containers/pytorch_extending_our_containers.ipynb
ChoiByungWook: done

advanced_functionality/pytorch_extending_our_containers/pytorch_extending_our_containers.ipynb
ChoiByungWook: Yes, this notebook should be updated when the TF and MXNet containers can handle this change.

advanced_functionality/pytorch_extending_our_containers/pytorch_extending_our_containers.ipynb
ChoiByungWook: Good point!

advanced_functionality/pytorch_extending_our_containers/pytorch_extending_our_containers.ipynb
ChoiByungWook: Added!

advanced_functionality/pytorch_extending_our_containers/pytorch_extending_our_containers.ipynb
djarpin: I worry "add additional functionality or configure your container" is too broad.  Specific examples.  E.g. "If you have a specific Python library like Pillow that we don't currently install but you always want included in your container."  etc.  This is kind of a minor point though.

PR #371, opened 2018-08-17
Added: MXNet Gluon CIFAR-10 Automatic Model Tuning vs random search

djarpin:


hyperparameter_tuning/mxnet_gluon_cifar10_random_search/cifar10.py
```diff
@@ -0,0 +1,186 @@
...
+
+    # load training and validation data
+    # we use the gluon.data.vision.CIFAR10 class because of its built in pre-processing logic,
+    # but point it at the location where SageMaker placed the data files, so it doesn't download them again.
+
+    part_index = 0
+    for i, host in enumerate(hosts):
+        if host == current_host:
+            part_index = i
+            break
```
mvsusp: What about 
```python
part_index = hosts.index(host)
```

djarpin: That does seem cleaner.  In this case, I copied `cifar10.py` right from the [existing example notebook](https://github.com/awslabs/amazon-sagemaker-examples/blob/master/sagemaker-python-sdk/mxnet_gluon_cifar10/cifar10.py#L41).  For consistency, I'll leave it as is.

PR #382, opened 2018-08-24
Update tensorflow notebooks to use sagemaker tensorflow container 1.9

yangaws:
In all tensorflow notebooks, explicitly pass framework version as argument. This is helpful because:

1. Users know which version they are using
2. We can freeze versions to older ones if new ones have bugs that cannot be fixed in short time

For now I tested all notebooks and pass version 1.9


yangaws:
@djarpin Oops not aware of those notebooks. I will test and update them later. Thanks!

PR #393, opened 2018-08-28
Fixed issue with setup.sh script

mattmcclean:
Previously running this script on a non-SageMaker notebook instance would give an error "Error: argument "agent" is wrong: table id value is invalid". Moved the variables inside the RUNNING_ON_NOTEBOOK_INSTANCE if statement to make sure it works outside SageMaker notebook instance

### DIFF
--- a/advanced_functionality/pytorch_extending_our_containers/utils/setup.sh
+++ b/advanced_functionality/pytorch_extending_our_containers/utils/setup.sh
@@ -41,16 +41,16 @@ fi
 # Notebook instance Docker networking fixes
 RUNNING_ON_NOTEBOOK_INSTANCE=`sudo iptables -S OUTPUT -t nat | grep -c 169.254.0.2`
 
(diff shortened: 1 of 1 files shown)


winstonaws:
Hey Matt, this script should only be needed on SageMaker notebook instances. It's also temporary and we're working to get rid of it, so I'd prefer not to make changes to it.

Is there a reason you need to run it outside of SageMaker notebooks? You should be able to use local mode without it, as long as you have docker and docker-compose installed.

winstonaws:
I'm going to close this PR given the above, but please reopen if you feel it's still needed and want to continue talking about it.

mattmcclean:
One reason for running outside of a SageMaker notebook is the limited disk space available for Docker images. Had to spin up an EC2 instance with the Deep Learning AMI instead.

winstonaws:
Right, what I mean is that you shouldn't need the setup.sh script when using local mode outside of SageMaker notebook instances. Did you have issues without the script?

PR #394, opened 2018-08-30
Update KMS encryption notebook with steps for batch transformation

apacker:
This adds additional steps to this notebook for batch transformation.

- Updated test dataset file generation to include a dataset without labels for batch transform
- Added upload of the test dataset (sans labels) to S3
- Updated endpoint validation to do a single prediction
- Added section for batch prediction using batch transform

### DIFF
--- a/advanced_functionality/handling_kms_encrypted_data/handling_kms_encrypted_data.ipynb
+++ b/advanced_functionality/handling_kms_encrypted_data/handling_kms_encrypted_data.ipynb
@@ -137,12 +137,7 @@
     "    feature_names = boston['feature_names']\n",
     "    data = pd.DataFrame(X, columns=feature_names)\n",
     "    if include_labels:\n",
-    "        target = pd.DataFrame(y, columns={'MEDV'})\n",
-    "        data['MEDV'] = y\n",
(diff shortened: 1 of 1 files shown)


advanced_functionality/handling_kms_encrypted_data/handling_kms_encrypted_data.ipynb
```diff
@@ -466,87 +458,121 @@
...
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {
-    "collapsed": true
-   },
+   "metadata": {},
    "outputs": [],
    "source": [
-    "client.delete_endpoint(EndpointName=endpoint_name)"
+    "print(\"Downloading prediction object...\")\n",
```
tomfaulhaber: This cell should be rewritten as:

```python
print("Downloading prediction object...")
s3.download_file(Bucket=bucket,
                 Key=prefix + '/predict/' + test_no_labels_file + '.out',
                 Filename='./predictions.csv')

preds = np.loadtxt('predictions.csv')
print('\nMedian Absolute Percent Error (MdAPE) = ', np.median(np.abs(y_test - preds) / y_test))
```

PR #413, opened 2018-09-18
Add batch transform to image-classification notebook - Part 2

Aloha106:
This adds additional steps to this notebook for batch transformation. The batch section includes:

* Create a Batch job using part of the validation dataset as input
* After job completes, check batch prediction results and print some sample output.

Similar changes: https://github.com/awslabs/amazon-sagemaker-examples/pull/402


gurumurthys:
Please add slightly more details on the results of batch transform. Some of the predictions are incorrect mainly because the number of epochs to train is set to 2 and that may not be sufficient to get a good model

PR #417, opened 2018-09-20
Update knn notebook to use SDK-based uri

asadoughi:


asadoughi:
@djarpin I've updated the commit. Thanks.

PR #426, opened 2018-09-29
add pytorch HPO sample notebook

fandotli:


hyperparameter_tuning/pytorch_mnist/hpo_pytorch_mnist.ipynb
```diff
@@ -0,0 +1,369 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "tuner.fit({'training': inputs})"
+   ]
+  },
+  {
```
djarpin: Do you want to throw a `tuner.wait()` in here?  If users don't fully understand that tuner.fit() orchestrates everything in the background, and that they'll need to monitor the tuning job's progress separately, then they'll get an error when trying to run the next cell.  Including `.wait()` will prevent this.  It does mean their notebook kernel will be locked for ~20 minutes though, which might not be ideal.  I'll leave it up to you.  Thanks.

PR #437, opened 2018-10-08
Add transaction fraud detection using simulated payments dataset

elangovana:
Hi I have added a new SageMaker Sample for fraud detection using simulated transactions dataset. The algorithm it uses for now is XGBoost. 
**Note** . Currently XGBoost doesn't support AUCPR (arc under precision curve) for highly imbalanced dataset....


fandotli:
picking this up. hi, elangovana, since it's been a while, do you have updated version of the example? Two quick comments:
1. You are using boto3 in the sample notebook while using SageMaker python sdk will make the example simpler. You may want to consider using it or at least mention it in the sample notebook.
2. it would be helpful to add more comments in the sample notebook to make it more educational. 

Thanks.

elangovana:
Hi No problem. will fix this and update the notebook soon

fandotli:
Unless I've missed something... it seems that you've defined the regex for metrics but they weren't really used since you didn't do HPO in this example. I'd suggest not to include metric definition if they are not going to be used.

elangovana:
Outdated - New fraud detection service available

PR #438, opened 2018-10-09
Added intel optimized packages

preethivenkatesh:
Edited the pip install commands to install CPU optimized intel Python packages. Tested end-to-end process(training and prediction locally).

### DIFF
--- a/advanced_functionality/scikit_bring_your_own/container/Dockerfile
+++ b/advanced_functionality/scikit_bring_your_own/container/Dockerfile
@@ -12,17 +12,19 @@ RUN apt-get -y update && apt-get install -y --no-install-recommends \
          python \
          nginx \
          ca-certificates \
+		 libpython2.7 \
     && rm -rf /var/lib/apt/lists/*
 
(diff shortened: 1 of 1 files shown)


preethivenkatesh:
Hi Marcio,
Thanks for reviewing. Here are my answers
1. Yes, intel packages work on all instances that SageMaker supports (given that they are all CPU-supported)
2. You can find our benchmarking  for using intel optimized packages(numpy, scipy and sklearn) here, Also we have a few for AWS instances
(https://software.intel.com/en-us/distribution-for-python/benchmarks#amazon-elastic-compute-cloud). Please let me know if you need more information.

jesterhazy:
Preethi,

Thanks for submitting this PR and your interest in SageMaker!
 
Unfortunately, we are going to reject this PR. The optimized python packages look like a very interesting way to get additional performance out of CPU instances, but including them in this tutorial notebook adds extra complexity and blurs the focus of the scikit example. This seems like a great topic for a blog post, though!

PR #452, opened 2018-11-05
Fixing recommender example

NRauschmayr:
I changed a couple of issues in the Gluon recommender example. 
It has been using an unnecessary ```try except``` block within the training loop. I assume ```try except``` was in place to avoid the issue when the last batch did not have the required batch_size and as such ```net(user, item)``` would throw an error. However, ```try except``` does not work here, because MXNet's engine would crash and not recover. Consequently, the Python interpreter is stuck.
To avoid this issue, the Dataloader has to be modified. The argument ```last_batch='rollover'``` makes sure that the last batch will be rolled over to next iteration, if it is not large enough.

It also appears, that ```SparseMatrixDataset(gluon.data.Dataset)``` is unnecessary. You can just use ```gluon.data.ArrayDataset(nd.array(train_df['user'].values, dtype=np.float32),
                                nd.array(train_df['item'].values, dtype=np.float32),
                                nd.array(train_df['star_rating'].values, dtype=np.float32))```

I also added ```num_workers=4``` in the DataLoader and increased the batch_size to increase GPU utilization.

I also modified the model itself, since it was using the same Dense layer for ```user``` and ```item``` which can be bad for training performance.  Dropout was set to 0 which means there was no Dropout applied.

### DIFF
--- a/introduction_to_applying_machine_learning/gluon_recommender_system/gluon_recommender_system.ipynb
+++ b/introduction_to_applying_machine_learning/gluon_recommender_system/gluon_recommender_system.ipynb
@@ -372,7 +372,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "num_embeddings = 256\n",
+    "num_embeddings = 64\n",
     "\n",
     "net = MFBlock(max_users=customer_index.shape[0], \n",
     "              max_items=product_index.shape[0],\n",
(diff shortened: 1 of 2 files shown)


introduction_to_applying_machine_learning/gluon_recommender_system/gluon_recommender_system.ipynb
```diff
@@ -933,9 +880,7 @@
   {
```
djarpin: This model is producing a perfect correlation between multiple user 6 and user 7's reviews.

NRauschmayr: Thanks a lot for the review! I added the changes.

larroy:
Hi Natalie. Do you have more info on the "would crash and not recover?" I would like to avoid MXNet crashing in such situation if possible.

NRauschmayr:
Hi Pedro,

We had a user two weeks ago who reported low GPU utilization for this example: https://discuss.mxnet.io/t/improving-gpu-usage-on-public-sagemaker-mxnet-example/2145 

When I tested it, I observed that the utilization is not only low but it would also drop to 1% after the first epoch has finished, and it would be stuck at 1% for many minutes. So I would just exit the process, assuming the process is somehow stuck since the last batch has triggered an exception (rollover was not defined in the original version). I asked my colleagues, but no one really knew what was happening and whether this could be due to the exception or not. After I re-factored  the code I did not see this issue anymore. I re-did  the tests for the original example and realized that the process was not stuck but just extremely slow. See the following plot:

gpu

It seems that the original example was stuck in loading the test data (which took many minutes) and that's why GPU utilization went down.  I am sorry about the confusion.

PR #454, opened 2018-11-06
Add notebook for ONNX export with MXNet 1.3

laurenyu:
This notebook is largely based on the [MXNet MNIST one](https://github.com/awslabs/amazon-sagemaker-examples/blob/master/sagemaker-python-sdk/mxnet_mnist/mxnet_mnist.ipynb), but takes the approach of the [HPO MXNet notebook](https://github.com/awslabs/amazon-sagemaker-examples/blob/master/hyperparameter_tuning/mxnet_mnist/hpo_mxnet_mnist.ipynb), showcasing a specific aspect of training and then linking to other notebooks that show the rest of the workflow.

### DIFF
--- a/sagemaker-python-sdk/mxnet_onnx_export/mnist.py
+++ b/sagemaker-python-sdk/mxnet_onnx_export/mnist.py
@@ -1,0 +1,122 @@
+import argparse
+import gzip
+import json
+import logging
+import os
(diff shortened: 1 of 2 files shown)


icywang86rui:
Let's ask Eric to take a look at this as well.

PR #481, opened 2018-11-21
Xgboost region repair

jpbarto:
This pull request makes changes to the MNIST and Abalone XGBoost notebooks.  In both notebooks there is a variable named `region` in the initial code block.  The impression customers get is that this controls which region the notebook will work with but in many subsequent code blocks `region` is redefined or ignored as S3 and SageMaker clients are instantiated.  

This update modifies both notebooks to adopt the initial definition of `region` through every line of code.  This will prevent customers from getting confusing error messages about regionality when they modify the `region` value.

### DIFF
--- a/introduction_to_amazon_algorithms/xgboost_mnist/xgboost_mnist.ipynb
+++ b/introduction_to_amazon_algorithms/xgboost_mnist/xgboost_mnist.ipynb
@@ -49,7 +49,6 @@
    "cell_type": "code",
    "execution_count": null,
    "metadata": {
-    "collapsed": true,
     "isConfigCell": true
    },
    "outputs": [],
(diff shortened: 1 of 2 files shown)


introduction_to_amazon_algorithms/xgboost_mnist/xgboost_mnist.ipynb
```diff
@@ -67,8 +66,10 @@
     "role = get_execution_role()\n",
     "\n",
     "region = boto3.Session().region_name\n",
+    "region='eu-west-1'\n",
+    "bucket='jasbarto-ml-ireland-bucket'\n",
```
yijiezz: I think these two lines are user-specific, and should not be there

yijiezz:
Same fix was applied by PR https://github.com/awslabs/amazon-sagemaker-examples/pull/674.

Closing this issue.

PR #482, opened 2018-11-23
Modernize gluon mnist example

ThomasDelteil:
This example is using a deprecated way of saving and loading parameters in Gluon.
Updating to current recommended way.

### DIFF
--- a/sagemaker-python-sdk/mxnet_gluon_mnist/mnist.py
+++ b/sagemaker-python-sdk/mxnet_gluon_mnist/mnist.py
@@ -65,7 +65,9 @@ def train(current_host, channel_input_dirs, hyperparameters, hosts, num_gpus):
                 break
 
         train_data = train_data[start:end]
-
+    
+    net.hybridize()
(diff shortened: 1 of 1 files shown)


sagemaker-python-sdk/mxnet_gluon_mnist/mnist.py
```diff
@@ -103,13 +105,11 @@ def train(current_host, channel_input_dirs, hyperparameters, hosts, num_gpus):
 
 def save(net, model_dir):
     # save the model
-    y = net(mx.sym.var('data'))
-    y.save('%s/model.json' % model_dir)
-    net.collect_params().save('%s/model.params' % model_dir)
+    y = net.export('%s/model'% model_dir)
```
laurenyu: do we need to set `y` anymore or can we remove it here?

ThomasDelteil: good point

PR #511, opened 2018-12-05
Use GPU instance for Neo Image Classification

hcho3:
To match the behavior of SageMaker Image Classification, Neo should use GPU instance for Image Classification.

cc @wweic


hcho3:
Closing for now. Will need to have CPU/GPU variants of Neo inference containers

PR #577, opened 2019-01-21
Proposing a new sample Object Detection notebook using a public Birds dataset

markproy:
The current sample notebooks for object detection leverage only the Pascal VOC dataset.  While this is very helpful for those getting started with Amazon SageMaker's Object Detection algorithm, I believe it would be helpful to provide additional examples with other image datasets.  My proposed notebook uses CalTech's CUB 200-2011 dataset.  I am currently authoring a corresponding blog post.  

Comparing to the existing object detection sample notebooks, this one breaks some new ground:

- We provide an example of how to translate bounding box specifications when providing images to SageMaker's algorithm. You will see code for generating the train.lst and val.lst files used to create recordIO files.  Not all datasets will be annotation-ready for RecordIO creation.

- We demonstrate how to improve an object detection model by adding training images that are flipped horizontally (mirror images).

- We give you a notebook for experimenting with object detection challenges with an order of magnitude more classes (200 bird species, as opposed to the 20 categories used by Pascal VOC).

- We show how to chart the accuracy improvements that occur across the epochs of the training job.

With this pull request, I'm looking for your feedback.  I have run the notebook through the mead notebook tester.  I left the output of the cells in the notebook so you could see what it would be like without having to wait for the full execution.  I recognize that on final submission, I would need to clear the output.

Thanks in advance for your feedback.
Mark


fandotli:
•	MXnet ->MXNet
•	“This role is used to give SageMaker access to your data in S3 will automatically be obtained from the role used to start the notebook.” -> “Here we will use the execution role the current notebook instance was given when it was created. This role has necessary permissions include access to your data in S3.”
•	I would recommend to move the code for retrieving training image to where the training job is created.
•	The data download part is currently commented out, which I assume need to be there.
•	In “define hyperparameters”, I would recommend to add a brief intro of automatic model tuning, and state that one can use it to search for the best hyperparameter combination.
•	“Therefore it might be a few minutes before we start getting data logs for our training jobs.”->” Therefore it might be a few minutes before our training job logs show up in CloudWatch.”
•	“You could even clone the job right from the console, and tweak some of the parameters to generate a new model.”->” You could even clone the job right from the console, and tweak some of the parameters to generate a new training job.”
•	In “host the model”, it’ll be good to mention that one can use batch transform if the inference does not need to be real time. 
•	“Training is a prolonged and compute heavy job that require a different of compute and memory requirements that hosting typically do not. “ – do not -> does not
•	Visualize_detection and show_bird_detection use different default thresholds (0.6 and 0.3 respectively), is there a particular reason for that? Better mention it in the comment
•	Add a comment in “clean up” to emphasize that the endpoint will keep running and cost customers money, so it needs to be deleted at the end.

PR #625, opened 2019-02-14
Update setup.sh file to skip if the user doesn't have root access

neelamgehlot:
### Description
* This change checks if current user has sudo access or not. If the
current user has sudo access then script is run as before and if
the current user doesn't have sudo access then skip the script.

### Testing
* Run setup.sh file from Notebook with the change with and without root access
* With root access
```
The user has root access.
SageMaker instance route table setup is ok. We are good to go.
SageMaker instance routing for Docker is ok. We are good to go!
```

* Without root access
```
sudo: a password is required
The user does not have root access. Everything required to run the notebook is already installed and setup. We are good to go!
```

### DIFF
--- a/advanced_functionality/pytorch_extending_our_containers/utils/setup.sh
+++ b/advanced_functionality/pytorch_extending_our_containers/utils/setup.sh
@@ -1,5 +1,13 @@
 #!/bin/bash
 
+sudo -n true
+if [ $? -eq 0 ]; then
+  echo "The user has root access."
+else
+  echo "The user does not have root access. Everything required to run the notebook is already installed and setup. We are good to go!"
+  exit 0
+fi
+
 # Do we have GPU support?
 nvidia-smi > /dev/null 2>&1
 if [ $? -eq 0 ]; then
--- a/advanced_functionality/tensorflow_bring_your_own/utils/setup.sh
+++ b/advanced_functionality/tensorflow_bring_your_own/utils/setup.sh
@@ -1,5 +1,13 @@
 #!/bin/bash
 
+sudo -n true
+if [ $? -eq 0 ]; then
+  echo "The user has root access."
+else
+  echo "The user does not have root access. Everything required to run the notebook is already installed and setup. We are good to go!"
+  exit 0
+fi
+
 # Do we have GPU support?
 nvidia-smi > /dev/null 2>&1
 if [ $? -eq 0 ]; then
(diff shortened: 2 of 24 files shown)


advanced_functionality/pytorch_extending_our_containers/utils/setup.sh
```diff
@@ -1,5 +1,13 @@
 #!/bin/bash
 
+sudo -n true
+if [ $? -eq 0 ]; then
+  echo "The user has root access."
+else
+  echo "The user does not have root access. Everything required to run the notebook is already installed and setup. We are good to go!"
```
nadiaya: Where comes from the assumption that if user doesn't have root access then everything is already installed and setup?

jaipreet-s: +1

This check should also be moved below the `Do we have GPU support` check. We don't need to fail in the absence of root access if the dependencies are already present.

cpranav: @nadiaya, 
Previously we had a long term bug on the services side around more than a year which resulted in race condition where sometime required setup was already installed, sometimes not. Now we have a fix for it going along with removing of root permissions. It's a prerequisite to remove root permissions as some of the things which we are doing in the script are impossible to do without having sudo permissions. That's from where assumption is coming up. 

Secondly as the script is idempotent if the root user runs this script multiple times it has no effect but for root free user this script cannot be run. We just cannot remove setup.sh script for one of the reason is sample notebooks executed outside of SageMaker won't be able to execute the notebook which requires setup.sh file.

@jaipreet-s I don't think we are failing in absence of root access, its an `exit 0` code.

jaipreet-s:
@cpranav Sure, fair point on not failing. However, this check still isn't needed if `Do we have GPU support` returns true, since the commands that use `sudo` are only used if that check fails

cpranav:
@jaipreet-s The code does not needs to be run even if its a CPU or GPU instance. If you scroll down half way in the file we can we have some parts in the script which runs irrespective of GPU or CPU like networking configuration and which requires `sudo` permissions.

jaipreet-s:
Ah, I see. Makes sense. nadiaya@'s point is still valid though - Just because the machine where the notebook is running does *not* have root access does *not* guarantee that everything is configured correctly.

PR #687, opened 2019-03-21
Strip out special characters from dataset row before converting token to int

neelamgehlot:
### Motivation
* Following error occurred during training job
```
Invoking script with the following command:
/usr/bin/python -m sentiment --batch-size 8 --embedding-size 50 --epochs 2 --learning-rate 0.01 --log-interval 1000

Traceback (most recent call last):
File "/usr/lib/python2.7/runpy.py", line 174, in _run_module_as_main
"__main__", fname, loader, pkg_name)
File "/usr/lib/python2.7/runpy.py", line 72, in _run_code
exec code in run_globals
File "/opt/ml/code/sentiment.py", line 338, in <module>
args.batch_size, args.epochs, args.learning_rate, args.log_interval, args.embedding_size)
File "/opt/ml/code/sentiment.py", line 37, in train
train_sentences, train_labels, _ = get_dataset(training_dir + '/train')
File "/opt/ml/code/sentiment.py", line 255, in get_dataset
label = int(tokens[0])
ValueError: invalid literal for int() with base 10: '500:'
```

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/sagemaker-python-sdk/mxnet_gluon_sentiment/sentiment.py
+++ b/sagemaker-python-sdk/mxnet_gluon_sentiment/sentiment.py
@@ -252,7 +252,7 @@ def get_dataset(filename):
     with open(filename) as f:
         for line in f:
             tokens = line.split()
-            label = int(tokens[0])
+            label = int(tokens[0].replace(':', '').replace(',', '').strip())
             words = tokens[1:]
             max_length = max(max_length, len(words))
             labels.append(label)
--- a/sagemaker-python-sdk/mxnet_gluon_sentiment/sentiment.py
+++ b/sagemaker-python-sdk/mxnet_gluon_sentiment/sentiment.py
@@ -252,6 +252,7 @@ def get_dataset(filename):
     with open(filename) as f:
         for line in f:
             tokens = line.split()
+            # Clean the data by stripping out special characters before converting label to int
             label = int(tokens[0].replace(':', '').replace(',', '').strip())
             words = tokens[1:]
             max_length = max(max_length, len(words))


sagemaker-python-sdk/mxnet_gluon_sentiment/sentiment.py
```diff
@@ -252,7 +252,7 @@ def get_dataset(filename):
     with open(filename) as f:
         for line in f:
             tokens = line.split()
-            label = int(tokens[0])
+            label = int(tokens[0].replace(':', '').replace(',', '').strip())
```
yangaws: Hi I was wondering if you have an example data that contains ':' here. I can understand we remove ',' so that '1,000' becomes '1000'. But I am worried for data like '100:2' which should be 50 and now becomes '1002'. 

I may be wrong. Just an example for ':' would be helpful to see what we should do.

neelamgehlot: I don't have an example data, the error mentioned in PR description is from training job logs.

Regarding your concern, in this case if you pass '100:2' to int() it would fail too with ValueError.

yangaws: @neelamgehlot 

As we talked offline last time. Could you add a comment on this line indicating that you are doing data cleaning. Then we can go ahead to merge the codes.

neelamgehlot: Added a comment in new commit

laurenyu:
given the lack of activity on this PR, I'm inclined to close it. Feel free to reopen if there is work to be added.

The training data is formatted as space-separated values with the label in the first column and words after that. I downloaded the training data locally and verified that the first column always is either a 0 or a 1:

```bash
$ cut -d " " -f 1 train | sort -u
0
1
```

the `int()` call that's referenced in the stacktrace is performed on `tokens[0]`, so it should only ever be calling `int()` on a `'0'` or `'1'`

PR #728, opened 2019-04-27
Fix error in building pytorch-fairseq container

smrmkt:
…ML error

*Issue #, if available:* 727

*Description of changes:* Added ingore-installed option to suppress can't uninstall distutil PyYAML


JonathanTaws:
Thanks @smrmkt for making this change. While in the current state this warning does not seem to break the creation of the container, I agree this is the right solution. Curious to know what change made PyYAML available by default, as this warning was not triggered before. 

Can you add the log trace of the error for completeness?

PR #777, opened 2019-06-24
fixes #775

### LINKED ISSUE #775: reinforcement learning/TSP order count does not match expected
https://github.com/awslabs/amazon-sagemaker-examples/blob/2469856b5734d6652eaeea844e1516ea2c83684e/reinforcement_learning/rl_traveling_salesman_vehicle_routing_coach/src/TSP_env.py#L98

expected:
this is supposed to produce a number of random points, defined in the class by `n_orders=`

observed:
there is a check later in the flow to ensure the order points are not placed on the restaurant point.  However, there is _not_ a check to ensure order points are not placed on top of previous order points.  This means that many episodes are training with fewer than  the desired number of `n_orders` points.  The episodes with fewer orders will always seem to outperform those with the correct number of orders, but its only because they have fewer orders to satisfy.

ezeeetm:
fixes #775

https://github.com/awslabs/amazon-sagemaker-examples/issues/775#issue-459509312


ezeeetm:
> Suggesting a minor change for efficiency:
@bbalaji-ucsd added.

PR #793, opened 2019-07-04
Edited Sklearn RF proposal with container parametrized path to support more regions

la-cruche:
*Issue #, if available:*

*Description of changes:*
following https://github.com/awslabs/amazon-sagemaker-examples/pull/792/, added a container ARN building function for the notebook to run in multiple regions. This PR replaces [the previous one](https://github.com/awslabs/amazon-sagemaker-examples/pull/792/)

### DIFF
--- a/sagemaker-python-sdk/scikit_learn_randomforest/Sklearn_on_SageMaker_end2end.ipynb
+++ b/sagemaker-python-sdk/scikit_learn_randomforest/Sklearn_on_SageMaker_end2end.ipynb
@@ -1,0 +1,669 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Develop, Train, Optimize and Deploy Scikit-Learn Random Forest\n",
+    "\n",
+    "* Doc https://sagemaker.readthedocs.io/en/stable/using_sklearn.html\n",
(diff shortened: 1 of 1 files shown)


sagemaker-python-sdk/scikit_learn_randomforest/Sklearn_on_SageMaker_end2end.ipynb
```diff
@@ -0,0 +1,669 @@
+{
```
laurencer: <p>Can this not use the map that is maintained in the Python SDK? Otherwise this will have to be updated for every new region launch correct?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/awslabs/amazon-sagemaker-examples/pull/793/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='19'/>

mvsusp: The map is only being used for the boto3 only example, I believe.

la-cruche: yes the map is only for the boto3 example. Goal is to show how to run sklearn training in plain boto3, eg something one would do in lambda

PR #820, opened 2019-07-24
New simapp support

sunil19m:
Root Cause:
  When creating the training image, the docker container uses pip to
  install rl-coach slim. When the pip command is executed rl_coach
  installs its dependencies. One of these dependencies is futures
  whose latest version is no longer supported by python 3, when the
  dependency is pulled in the build fails because pip exits the shell
  with a non-zero return code.

Solution:
  Set the version of futures to the minimum required version which
  is 3.1.1, this prevents pip from pulling in the non-compatible version.

  As part of this revert the upgrade of rl coach which appears to have
  an issue when reading the checkpoint from disk, this causes bad
  training jobs.

Why and When Introduced:
  Introduced at implementation, the developer couldn't have known that
  futures would eventually not be supported by python3.

Where Else:
  Looks like we version futures in the sim app so this bug will not impact
  the robomaker sid eof things.

Testing:
  Validated build and validated that training correctly happens. Validated
  that the sim app works as expected.


vrkhare:
new-simapp-support has newer pandas

PR #856, opened 2019-08-23
Add an XGBoost script mode example

iyerr3:
*Issue #, if available:*

*Description of changes:*
- Add a new notebook describing script mode for new XGBoost framework container
- Add a new accompanying script file for above notebook


eslesar-aws:
@iyerr3 Only comment is that typically notebooks have a cell (either !cat or !pygmentize) to display the training script.

iyerr3:
@eslesar-aws Added the training script in the notebook

ericangelokim:
* Introduction needs an intro to xgboost, for people not familiar with xgboost, as well as a more detailed introduction to what script mode training is
* Introduction should also probably reference the older xgboost algorithm mode notebooks as 'alternative ways of running training', for customers who are new to SageMaker
* We use this same dataset in other notebooks right? If not this will require approval from data hosts
* The script we use will benefit from an explanation of rabit is, since it is referenced in the script.
* We want notebooks to be end to end, in other words, include inference to prove to customers correctness.

I would sync with a PM about this notebook and the script. How urgent is this?

PR #857, opened 2019-08-23
Add Contextual Bandits example

sidd1809:
*Description of changes:*

This commit adds an example that demonstrates how SageMaker can be used for Contextual Bandits


vrkhare:
New sagemaker rl feature on multi-arm contextual bandits

PR #887, opened 2019-09-20
Fix: set job_name_prefix, fix job_name

orcaman:
This notebook doesn't run because it seems to be using an undeclared variable, `job_name_prefix`. This PR adds the missing job_name_prefix and sets `job_name` using the new `job_name_prefix` variable.


laurenyu:
looks like this has since been fixed

PR #890, opened 2019-09-26
Fix HVAC EnergyPlus RL Hosting

bbalaji-ucsd:
*Issue #, if available:*

*Description of changes:*
The hosting endpoint was using Python2 by default and failed stochastically. Fixed the issue by updating to use RLEstimator deploy module instead.


annaluo676:
Looks good to me.

PR #919, opened 2019-11-07
Support restoring checkpoint in ray; Add packages in dockerfile

annaluo676:
*Description of changes:*
1. allow experiment restoration from the CHECKPOINT channel
2. enable custmerized checkpoint frequency
3. add packeages in dockerfile for efficency & memory monitoring

### DIFF
--- a/reinforcement_learning/common/sagemaker_rl/ray_launcher.py
+++ b/reinforcement_learning/common/sagemaker_rl/ray_launcher.py
@@ -98,14 +98,7 @@ class SageMakerRayLauncher(object):
         # TODO: move this to before customer-specified so they can override
         hyperparams_dict["rl.training.local_dir"] = INTERMEDIATE_DIR
         hyperparams_dict["rl.training.checkpoint_at_end"] = True
-
-        try:
-            customized_checkpoint_frequency = config['training']['checkpoint_freq']
-            hyperparams_dict["rl.training.checkpoint_freq"] = customized_checkpoint_frequency
-        except KeyError:
-            print("Using default checkpoint frequency: 10")
-            hyperparams_dict["rl.training.checkpoint_freq"] = 10
-            pass
+        hyperparams_dict["rl.training.checkpoint_freq"] = config['training'].get('checkpoint_freq', 10)
 
         self.hyperparameters = ConfigurationList()  # TODO: move to shared
         for name, value in hyperparams_dict.items():
@@ -244,24 +237,28 @@ class SageMakerRayLauncher(object):
     def set_up_checkpoint(self, config=None):
         try:
             checkpoint_dir = config['training']['restore']
-            print("Found checkpoint dir in user config.")
+            print("Found checkpoint dir %s in user config." %checkpoint_dir)
+            return config
         except KeyError:
             pass
 
         try:
             # validate files in checkpoint channel and set restore dir in ray config
             validation = 0
-            if len(os.listdir(CHECKPOINT_DIR)) is not (2 or 3):
+            num_checkpoint_files = len(os.listdir(CHECKPOINT_DIR))
+                
+            if num_checkpoint_files is not (2 or 3):
                 raise RuntimeError("Unexpected files in checkpoint dir.",
-                                   "Please check ray documents for the correct checkpoint format.")
-            checkpoint_dir = ""
+                                    "Please check ray documents for the correct checkpoint format.")
+
+            checkpoint_file_in_container = ""
             for filename in os.listdir(CHECKPOINT_DIR):
                 is_tune_metadata= filename.endswith("tune_metadata")
                 is_extra_data = filename.endswith("extra_data")
                 is_checkpoint_meta = is_tune_metadata + is_extra_data
                 validation += is_checkpoint_meta
                 if not is_checkpoint_meta:
-                    checkpoint_dir = os.path.join(CHECKPOINT_DIR, filename)
+                    checkpoint_file_in_container = os.path.join(CHECKPOINT_DIR, filename)
 
             if ray.__version__ >= "0.6.5":
                 if validation is not 1:
@@ -269,18 +266,17 @@ class SageMakerRayLauncher(object):
             else:
                 if validation is not 2:
                     raise RuntimeError("Failed to find .tune_metadata or .extra_data to restore checkpoint")
-            if checkpoint_dir:
-                print("Found checkpoint: %s. Setting restore path in ray config." %checkpoint_dir)
-                print( "Important! Ray with version <=7.2 may report \"Did not find checkpoint file\" even if the",
-                    "experiment is actually restored successfully. Please check \"training_iteration\" in the experiment",
-                    "info to confirm."
-                    )
-                config['training']['restore'] = checkpoint_dir
+                    
+            if checkpoint_file_in_container:
+                print("Found checkpoint: %s. Setting `restore` path in ray config." %checkpoint_file_in_container)
+                config['training']['restore'] = checkpoint_file_in_container
             else:
-                print("No checkpoint found in %s. Training from scratch." %CHECKPOINT_DIR)
+                print("No valid checkpoint found in %s. Training from scratch." %CHECKPOINT_DIR)
         except OSError:
             print("No checkpoint path specified. Training from scratch.")
             pass
+            
+        return config
 
     def launch(self):
         """Actual entry point into the class instance where everything happens.
@@ -297,9 +293,13 @@ class SageMakerRayLauncher(object):
         ray.init(**ray_cluster_config)
         experiment_config = self.get_experiment_config()
         experiment_config = self.customize_experiment_config(experiment_config)
-        self.set_up_checkpoint(experiment_config)
+        experiment_config = self.set_up_checkpoint(experiment_config)
         
         print("Running experiment with config %s" % json.dumps(experiment_config, indent=2))
+        print("Important! Ray with version <=7.2 may report \"Did not find checkpoint file\" even if the",
+              "experiment is actually restored successfully. If restoration is expected, please check",
+              "\"training_iteration\" in the experiment info to confirm."
+             )
         run_experiments(experiment_config)
         all_wokers_host_names = self.get_all_host_names()[1:]
         # If distributed job, send TERMINATION_SIGNAL to all workers.
--- a/reinforcement_learning/common/sagemaker_rl/ray_launcher.py
+++ b/reinforcement_learning/common/sagemaker_rl/ray_launcher.py
@@ -203,8 +203,8 @@ class SageMakerRayLauncher(object):
             _, ext = os.path.splitext(source_path)
             destination_path = os.path.join(MODEL_OUTPUT_DIR, "checkpoint%s" % ext)
             copyfile(source_path, destination_path)
-            print("Saved the checkpoint file %s as %s" % (source_path, destination_path))     
-           
+            print("Saved the checkpoint file %s as %s" % (source_path, destination_path))
+
     def save_experiment_config(self, config):
         with open(os.path.join(MODEL_OUTPUT_DIR, "params.json"), "w") as f:
             json.dump(config, f, indent=2)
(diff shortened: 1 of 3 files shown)


annaluo676:
Tested on rl-roboschool-reacher example with local_mode True and False, on local mac.

reinforcement_learning/common/sagemaker_rl/ray_launcher.py
```diff
@@ -97,7 +98,15 @@ def customize_experiment_config(self, config):
         # TODO: move this to before customer-specified so they can override
         hyperparams_dict["rl.training.local_dir"] = INTERMEDIATE_DIR
         hyperparams_dict["rl.training.checkpoint_at_end"] = True
-        hyperparams_dict["rl.training.checkpoint_freq"] = 10
+
+        try:
+            customized_checkpoint_frequency = config['training']['checkpoint_freq']
```
yijiezz: You can use `customized_checkpoint_frequency = config['training'].get('checkpoint_freq', 10)` to set default value for `checkpoint_freq `. In that way I guess you can bypass the `try except`

reinforcement_learning/common/sagemaker_rl/ray_launcher.py
```diff
@@ -231,6 +240,47 @@ def save_checkpoint_and_serving_model(self, algorithm=None, env_string=None, con
         # To ensure SageMaker local mode works fine
         change_permissions_recursive(INTERMEDIATE_DIR, 0o777)
         change_permissions_recursive(MODEL_OUTPUT_DIR, 0o777)
+        
+    def set_up_checkpoint(self, config=None):
+        try:
+            checkpoint_dir = config['training']['restore']
+            print("Found checkpoint dir in user config.")
+        except KeyError:
```
yijiezz: Same here, you can `.get` to avoid `KeyError` Also not quite understand this line, it seems like no matter what happened here, `checkpoint_dir ` will be reset into "" in line 257

reinforcement_learning/common/sagemaker_rl/ray_launcher.py
```diff
@@ -231,6 +240,47 @@ def save_checkpoint_and_serving_model(self, algorithm=None, env_string=None, con
...
+            pass
+
+        try:
+            # validate files in checkpoint channel and set restore dir in ray config
+            validation = 0
+            if len(os.listdir(CHECKPOINT_DIR)) is not (2 or 3):
+                raise RuntimeError("Unexpected files in checkpoint dir.",
+                                   "Please check ray documents for the correct checkpoint format.")
+            checkpoint_dir = ""
```
yijiezz: A little bit confusing, is `CHECKPOINT_DIR ` the directory for checkpoint or `checkpoint_dir`? Seems `checkpoint_dir` is the file path of checkpoint?

reinforcement_learning/common/sagemaker_rl/ray_launcher.py
```diff
@@ -231,6 +240,47 @@ def save_checkpoint_and_serving_model(self, algorithm=None, env_string=None, con
...
+            for filename in os.listdir(CHECKPOINT_DIR):
+                is_tune_metadata= filename.endswith("tune_metadata")
+                is_extra_data = filename.endswith("extra_data")
+                is_checkpoint_meta = is_tune_metadata + is_extra_data
+                validation += is_checkpoint_meta
+                if not is_checkpoint_meta:
+                    checkpoint_dir = os.path.join(CHECKPOINT_DIR, filename)
```
yijiezz: If there are multiple files under `CHECKPOINT_DIR `, `checkpoint_dir ` will be override multiple times?

reinforcement_learning/common/sagemaker_rl/ray_launcher.py
```diff
@@ -231,6 +240,47 @@ def save_checkpoint_and_serving_model(self, algorithm=None, env_string=None, con
...
+            if len(os.listdir(CHECKPOINT_DIR)) is not (2 or 3):
+                raise RuntimeError("Unexpected files in checkpoint dir.",
+                                   "Please check ray documents for the correct checkpoint format.")
+            checkpoint_dir = ""
+            for filename in os.listdir(CHECKPOINT_DIR):
+                is_tune_metadata= filename.endswith("tune_metadata")
+                is_extra_data = filename.endswith("extra_data")
+                is_checkpoint_meta = is_tune_metadata + is_extra_data
+                validation += is_checkpoint_meta
+                if not is_checkpoint_meta:
```
yijiezz: Once come into a valid checkpoint file, this condition will never fail after that

reinforcement_learning/common/sagemaker_rl/ray_launcher.py
```diff
@@ -231,6 +240,47 @@ def save_checkpoint_and_serving_model(self, algorithm=None, env_string=None, con
...
+            else:
+                if validation is not 2:
+                    raise RuntimeError("Failed to find .tune_metadata or .extra_data to restore checkpoint")
+            if checkpoint_dir:
+                print("Found checkpoint: %s. Setting restore path in ray config." %checkpoint_dir)
+                print( "Important! Ray with version <=7.2 may report \"Did not find checkpoint file\" even if the",
+                    "experiment is actually restored successfully. Please check \"training_iteration\" in the experiment",
+                    "info to confirm."
+                    )
+                config['training']['restore'] = checkpoint_dir
```
yijiezz: This will override what customers put in the config restore path?

reinforcement_learning/rl_roboschool_ray/Dockerfile
```diff
@@ -29,7 +29,10 @@ RUN curl -fSsL -O https://bootstrap.pypa.io/get-pip.py && \
 
 RUN pip install --upgrade \
     pip \
-    setuptools
+    setuptools \
```
yijiezz: Does the current Docker file not work?

annaluo676: The current Docker file works, this is for better user experience. `setproctitle` and `psutil` are for Ray dashboard and memory tracking; lz4 is sample compression which impacts RLlib's performance.

reinforcement_learning/common/sagemaker_rl/ray_launcher.py
annaluo676: Yeah, the transition from downloading the checkpoint files to how Ray can actually use them is a bit tricky. When customer provides a checkpoint s3 location, say, `s3://path/to/ray/results/checkpoint_100`, the files under `checkpoint_100` get downloaded to `CHECKPOINT_DIR`.  There will only be 2 files (`checkpoint-100` and `checkpoint-100.tune_metadata`) for ray version >= 0.6.5, and 3 files (another `checkpoint-100.extra_data`) for the older version. 

Then for ray to recognize them, `restore` should be set to `CHECKPOINT_DIR/checkpoint-100`, instead of `CHECKPOINT_DIR`. The logic here is to extrat `checkpoint-100` and attach it to `CHECKPOINT_DIR`

reinforcement_learning/common/sagemaker_rl/ray_launcher.py
annaluo676: As explained above, there will only be at most 3 files. So `checkpoint_dir` wouldn't be overwritten.

reinforcement_learning/common/sagemaker_rl/ray_launcher.py
annaluo676: Customer can restore the experiment in two ways: directly specifying the `restore` config in training script, or putting the checkpoint files in CHECKPOINT channel and let SageMaker take care of it. There is an edge case where customer provides both and I'm testing that.

reinforcement_learning/common/sagemaker_rl/ray_launcher.py
annaluo676: Good point.

reinforcement_learning/common/sagemaker_rl/ray_launcher.py
```diff
@@ -231,6 +233,50 @@ def save_checkpoint_and_serving_model(self, algorithm=None, env_string=None, con
...
+        try:
+            # validate files in checkpoint channel and set restore dir in ray config
+            validation = 0
+            num_checkpoint_files = len(os.listdir(CHECKPOINT_DIR))
+                
+            if num_checkpoint_files is not (2 or 3):
+                raise RuntimeError("Unexpected files in checkpoint dir.",
+                                    "Please check ray documents for the correct checkpoint format.")
+
+            checkpoint_file_in_container = ""
```
yijiezz: Customer has full control on what to upload to 'checkpoint' channel (e.g. one .tune_metadata file with two random files can bypass these validation). It will be good to have more thorough validations.

reinforcement_learning/common/sagemaker_rl/ray_launcher.py
```diff
@@ -231,6 +233,50 @@ def save_checkpoint_and_serving_model(self, algorithm=None, env_string=None, con
...
+            else:
+                if validation is not 2:
+                    raise RuntimeError("Failed to find .tune_metadata or .extra_data to restore checkpoint")
+                    
+            if checkpoint_file_in_container:
+                print("Found checkpoint: %s. Setting `restore` path in ray config." %checkpoint_file_in_container)
+                config['training']['restore'] = checkpoint_file_in_container
+            else:
+                print("No valid checkpoint found in %s. Training from scratch." %CHECKPOINT_DIR)
+        except OSError:
```
yijiezz: which line will possiblly throw OSError ?

annaluo676: `os.listdir()`

reinforcement_learning/common/sagemaker_rl/ray_launcher.py
annaluo676: There is a tradeoff in terms of validation -- how much would SageMaker RL handle, and how much left for Ray. In case where customer points 'checkpoint' channel to a ray model checkpoint on s3, Ray takes care of the validation e2e. It's absolutely possible that customer put some random files there, and there's the reason why the number of files is checked by SageMaker. I think after that level of validation, the rest can be handled to Ray. Error message such as `cannot find .tune_metadata` should be clear to users.

reinforcement_learning/common/sagemaker_rl/ray_launcher.py
```diff
@@ -201,8 +203,8 @@ def copy_checkpoints_to_model_output(self):
             _, ext = os.path.splitext(source_path)
             destination_path = os.path.join(MODEL_OUTPUT_DIR, "checkpoint%s" % ext)
             copyfile(source_path, destination_path)
-            print("Saved the checkpoint file %s as %s" % (source_path, destination_path))
-
+            print("Saved the checkpoint file %s as %s" % (source_path, destination_path))     
+
```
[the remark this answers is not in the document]
annaluo676: addressed

reinforcement_learning/common/sagemaker_rl/ray_launcher.py
```diff
@@ -231,6 +233,50 @@ def save_checkpoint_and_serving_model(self, algorithm=None, env_string=None, con
...
+            return config
+        except KeyError:
+            pass
+
+        try:
+            # validate files in checkpoint channel and set restore dir in ray config
+            validation = 0
+            num_checkpoint_files = len(os.listdir(CHECKPOINT_DIR))
+                
+            if num_checkpoint_files is not (2 or 3):
```
saurabh3949: (2 or 3) will always resolve to 2. You can change to 
if num_checkpoint_files is not in [2,3]

reinforcement_learning/common/sagemaker_rl/ray_launcher.py
```diff
@@ -231,6 +233,50 @@ def save_checkpoint_and_serving_model(self, algorithm=None, env_string=None, con
...
+                raise RuntimeError("Unexpected files in checkpoint dir.",
+                                    "Please check ray documents for the correct checkpoint format.")
+
+            checkpoint_file_in_container = ""
+            for filename in os.listdir(CHECKPOINT_DIR):
```
saurabh3949: Should we do a recursive search for the checkpoint files? Customer can have nested directories.

annaluo676: Added

PR #980, opened 2020-01-13
Adopting to the new format of the tuning job status

guyernest:
*Issue #, if available:*
The example notebook is not working any more due to the changes of the API
*Description of changes:*
Adding the name of the tuning job as the key for the results


aws-rbs (commented):
I a getting an error when running the second code block:
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-13-93475d73ebe0> in <module>
      9 print("%d training jobs have completed" % job_count)
     10 
---> 11 is_minimize = (tuning_job_result['TrainingJobDefinitions'][0]['TuningObjective']['Type'] != 'Maximize')
     12 objective_name = tuning_job_result['TrainingJobDefinitions'][0]['TuningObjective']['MetricName']

KeyError: 'TrainingJobDefinitions'

aws-rbs:
But the existing version is broken too.

bobbruno:
I can't open the logs, and I find it weird that it's even returning something and not raising an exception on the `describe_hyper_parameter_tuning_job` call. The test build is probably just passing `'YOUR-HYPERPARAMETER-TUNING-JOB-NAME'` as the `HyperParameterTuningJobName` parameter, and I doubt an HPO job with that name exists. I'd expect the call to raise a `ResourceNotFound` exception which should come back to you, not what looks like an empty `tuning_job_result` dictionary, which seems to be the cause for the `KeyError`.

Fixing that requires that the `tuning_job_name` variable contains an existing HPO job name when the test build is run. If you can agree with the CI/CD team on an existing tuning job name and leave that as the default on the notebook, you should be able to pass the test.

aaronmarkham:
There's no easy fix for this to pass CI. I'm going to go ahead and merge.

aws-rbs:
I testing it with a tuning job I had but the fix still failed as noted. So this will not succeed even if you give it a job.

PR #983, opened 2020-01-14
Supporting multirollout for the simapp released as of reInvent 2019

sunil19m:
*Issue #, if available:*
The launch of multiple sensors in the deepracer requires modification to notebook.
The simapp now supports the following

Multiple sensors - stereo, lidar, front camera. Along with multiple combination of these
Simulation environment to avoid obstacles in the training car
Different neural network. Shallow/Deep
Multi rollout

*Description of changes:*


ReHoss:
Hi, since a new object called MultiAgentGraphManager is introduced in this commit, is it still possible to use presets file to choose de RL Coach algorithm (e.g. DQN) ?  def write_frozen_graph does not exist anymore, what was the purpose of this function ? Except presets, which files should I modify ?

Thank you for your consideration

PR #1000, opened 2020-01-25
Raise if bash script hits an error

rahul003:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/advanced_functionality/distributed_tensorflow_mask_rcnn/prepare-s3-bucket.sh
+++ b/advanced_functionality/distributed_tensorflow_mask_rcnn/prepare-s3-bucket.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -e
+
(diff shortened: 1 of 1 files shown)


advanced_functionality/distributed_tensorflow_mask_rcnn/prepare-s3-bucket.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -e
```
chenliu0831: Could be a bit more strict: 

```
# Safer bash: -e 'error', -u 'undefined variable', -o 'pipefail'
set -euo pipefail
```

rahul003: That causes the script to fail

PR #1008, opened 2020-01-28
adding studio getting started

sojiadeshina:
introducing new features with sagemaker studio
*Description of changes:*
notebook walks through sagemeaker features like experiments, debugger and model monitor with the use case of training an xgboost model on the abalone dataset.


aws_sagemaker_studio/getting_started/xgboost_customer_churn.py
```diff
@@ -0,0 +1,138 @@
...
+        else None
+    )
+
+    hook = create_smdebug_hook(
+        out_dir=output_uri,
+        frequency=args.smdebug_frequency,
+        collections=collections,
+        train_data=dtrain,
+        validation_data=dval,
+    )
```
Vikas-kum: This is not required. The exisiting script should just work with ZCC support and loss not decreasing rule.

PR #1016, opened 2020-02-05
update studio getting started notebook

sojiadeshina:
adding extra trial runs with different hyperparameters for studio comparisons
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/aws_sagemaker_studio/getting_started/xgboost_customer_churn_studio.ipynb
+++ b/aws_sagemaker_studio/getting_started/xgboost_customer_churn_studio.ipynb
@@ -129,7 +129,7 @@
     "---\n",
     "## Train\n",
     "\n",
-    "Moving onto training! We will be training a class of models known as gradient boosted decision trees on the data we just uploaded. \n",
+    "Moving onto training! We will be training a class of models known as gradient boosted decision trees on the data we just uploaded using the XGBoost library. \n",
     "\n",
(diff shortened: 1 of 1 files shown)


aws_sagemaker_studio/getting_started/xgboost_customer_churn_studio.ipynb
```diff
@@ -637,8 +684,8 @@
    "source": [
     "# First, copy over some test scripts to the S3 bucket so that they can be used for pre and post processing\n",
     "code_prefix = '{}/code'.format(prefix)\n",
-    "S3Uploader.upload('preprocessor.py', 's3://{}/{}'.format(bucket,code_prefix))\n",
-    "S3Uploader.upload('postprocessor.py', 's3://{}/{}'.format(bucket,code_prefix))"
+    "pre_processor_script = S3Uploader.upload('preprocessor.py', 's3://{}/{}'.format(bucket,code_prefix))\n",
```
knakad: Quick note that you can also just pass the local path to the script directly to the ModelMonitor object.

sojiadeshina: but that puts it in the default s3 bucket no?

knakad: Correct, it'll put it in the default bucket and under the job name and a guid.  
Since you explicitly want all these files in the same S3 location, then you should keep doing this =)

PR #1045, opened 2020-02-13
Fix timeout error waiting for metrics

gabehollombe-aws:
This change fixes an error in the notebook. We were using `s3_bucket` to construct the `training_metrics_path` variable, but we should have been using `s3_prefix`. Without this change, this cell times out because it's looking for a file in the wrong path.

*Issue #, if available:*

*Description of changes:*


gabehollombe-aws:
Bump. Has anyone reviewed this yet?

PR #1052, opened 2020-02-19
[DO NOT MERGE] change: upgrade pipemode example framework version

otter-bunny:
*Issue #, if available:*

*Description of changes:*
Upgrade framework version for TensorFlow PipeMode example. Modified the pipemode script for tensorflow 2.0. (DO NOT MERGE until the images are released)


owen-t:
Also consider changing the pipemode.py script to be compatible with the TensorFlow container script mode.

ishaaq:
> Also consider changing the pipemode.py script to be compatible with the TensorFlow container script mode.

+1 - in its current form the training job will run without error however it won't actually exercise PipeModeDataset since nothing will actually call `_input_fn(...)` at run time.

Issue #1085, opened 2020-03-09
ground truth example fails to copy open-images-dataset


rudpot:
The ground_truth_labeling_jobs/from_unlabeled_data_to_deployed_machine_learning_model_ground_truth_demo_image_classification demo fails to copy data from the open-images-dataset because of signed requests.

Workaround is get/put copies.

TEChopra1000:
@rudpot can you copy and paste the error you saw? I was able to successfully download the data set, and upload it to S3. Did you check that the AWS role or user you were using has permission to copy to S3 - namely s3:GetObject and s3:PutObject?

natsirtguy:
I also could not reproduce the error @rudpot saw, `s3.copy` worked for me even though I didn't tell the client to use unsigned requests.

I'll close this for now, feel free to reopen @rudpot if the error still occurs for you.

PR #1086, opened 2020-03-09
Fix copy of image data to work without signature

rudpot:
As per the documentation on CVDF, downloads should happen with the no-sign-request option. As published the notebook will fail to copy the images with a 403 error from the source about HeadObject calls being forbidden. Suppressing signatures in the client using config=Config(signature_version=UNSIGNED) will fail at the destination end with an authorization failure.

It appears that the S3 COPY operation requires the authentication on both ends to be the same, even with providing a separate s3 client for the Head/Get operation. So the only plausible workaround appears to be to do a get/put copy as proposed by this patch.

*Issue #, if available: #1085

*Description of changes: added get/put workaround


natsirtguy:
I could not reproduce the error @rudpot saw, `s3.copy` worked for me even though I didn't tell the client to use unsigned requests.

I'll close this for now, feel free to reopen @rudpot if the error still occurs for you.

PR #1092, opened 2020-03-17
Add autopilot customer churn notebook

juliodelgadoaws:
Similar to the XGBoost customer churn notebook, but using autopilot
instead.

*Issue #, if available:*

*Description of changes:*

To support Aurora ML blog (https://aws.amazon.com/blogs/aws/new-for-amazon-aurora-use-machine-learning-directly-from-your-databases/) it is necessary to have a Customer Churn Autopilot example. 

The blog currently references: https://github.com/awslabs/amazon-sagemaker-examples/blob/master/introduction_to_applying_machine_learning/xgboost_customer_churn/xgboost_customer_churn.ipynb but the resulting endpoint from this notebook is not usable from Aurora, since it does not include all the pre/post processing code generated.

### DIFF
--- a/autopilot/autopilot_customer_churn.ipynb
+++ b/autopilot/autopilot_customer_churn.ipynb
@@ -241,7 +241,7 @@
     "      'DataSource': {\n",
     "        'S3DataSource': {\n",
     "          'S3DataType': 'S3Prefix',\n",
-    "          'S3Uri': 's3://{bucket}/{prefix}/train'.format(bucket,prefix)\n",
+    "          'S3Uri': 's3://{}/{}/train'.format(bucket,prefix)\n",
     "        }\n",
     "      },\n",
     "      'TargetAttributeName': 'Churn?'\n",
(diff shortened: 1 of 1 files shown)


pdasamzn:
I would restrict the MaxCandidates to be less that 20 with explicit instruction on how to increase. https://docs.aws.amazon.com/sagemaker/latest/APIReference/API_AutoMLJobCompletionCriteria.html#sagemaker-Type-AutoMLJobCompletionCriteria-MaxCandidates

This would imply that we may not have 96% accuracy as mentioned in the later section of the doc.

juliodelgadoaws:
> You can specify that the 'Churn?' contains 2 values & hence it's a binary classification problem. However, we don't need to specify as Autopilot will be able to infer it.

Make sense.

juliodelgadoaws:
> I would restrict the MaxCandidates to be less that 20 with explicit instruction on how to increase. https://docs.aws.amazon.com/sagemaker/latest/APIReference/API_AutoMLJobCompletionCriteria.html#sagemaker-Type-AutoMLJobCompletionCriteria-MaxCandidates
> 
> This would imply that we may not have 96% accuracy as mentioned in the later section of the doc.

Will modify and run a couple of times to make sure I have a correct ballpack value in the text for the accuracy.

pdasamzn:
lgtm!

PR #1118, opened 2020-03-31
Marketplace - New sample notebook

kwwaikar:
*Description of changes:*
Adding a Jupyter notebook that showcases how to use domain specific NER model from AWS Marketplace to make the financial transaction data processing efficient.


aws_marketplace/using_model_packages/financial_transaction_processing/src/model_package_arns.py
```diff
@@ -0,0 +1,22 @@
+
+class ModelPackageArnProvider:
+    
+    @staticmethod
+    def get_transactional_NER_model_package_arn(current_region):
+        mapping = {
```
yuanzhua, line 6: Is it by design only ten regions are supported?

kwwaikar: Hi @yuanzhua,

Looks like i missed this while updating the version number. The model supports 16 regions and i have updated the same in code in this commit - https://github.com/awslabs/amazon-sagemaker-examples/pull/1118/commits/591d8c856dad9cc0bd797318fee85a13c3d5d3aa

PR #1130, opened 2020-04-07
Added --no-check-certificate to wget

bustosjuan:
*Description of changes:*
archive.ics.uci.edu's certificate has expired, added --no-check-certificate to be able to download the file


aaronmarkham:
Looks like they renewed their cert, so this change isn't needed anymore. Please reopen if you don't agree. Thanks for the contribution.

PR #1137, opened 2020-04-13
Wait after performing some steps in the Studio Getting Started notebook

nigenda-amazon:
*Issue #, if available:*

*Description of changes:*
* Add wait after enabling data capture to give data upload time to take place
* Add wait on monitoring execution to allow it to complete
* Wait after deleting monitoring schedule, before proceeding to endpoint deletion


nigenda-amazon:
Forgot to add that I tested this by restarting the kernel and clicking "Run all" to ensure all cells now succeed when running in order from scratch (unlike the case prior to this commit).

PR #1158, opened 2020-04-22
Add missing read() calls to usages of invoke_endpoint()

kelleraa:
If the response body is not fully consumed with read(), then
the underlying connection will not be made available for reuse
(at least in a timely or predictable manner). This call was
missing in places where we don't actually care about the body
content.

### DIFF
--- a/reinforcement_learning/rl_game_server_autopilot/sagemaker/rl_gamerserver_ray.ipynb
+++ b/reinforcement_learning/rl_game_server_autopilot/sagemaker/rl_gamerserver_ray.ipynb
@@ -869,8 +869,8 @@
     "      ContentType=content_type,\n",
     "      Accept=accept,\n",
     "      Body=last_observations\n",
-    "    )",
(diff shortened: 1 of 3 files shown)


reinforcement_learning/rl_game_server_autopilot/sagemaker/rl_gamerserver_ray.ipynb
```diff
@@ -869,7 +869,8 @@
     "      ContentType=content_type,\n",
     "      Accept=accept,\n",
     "      Body=last_observations\n",
-    "    )"
+    "    )",
```
nigenda-amazon, line 872: Minor, should probably have trailing newline so that it doesn't look squished:
```
   "    )\n",
```

kelleraa: Fixed. Thanks!

PR #1166, opened 2020-04-24
Fix scikit-image version with the numpy version used in notebook

AdnanKukuljac:
*Issue #, if available:*

*Description of changes:*
Scikit-image is not compatible with the numpy version 1.16.4 which is used in SageMaker Notebook. More about this issue: https://github.com/scikit-image/scikit-image/issues/3649 This fix will install scikit-image 0.14.2 which works with numpy 1.16.4. Tested in us-west-2.


hyperparameter_tuning/mxnet_gluon_cifar10_random_search/hyperparameter_tuning_mxnet_gluon_cifar10_random_search.ipynb
```diff
@@ -41,6 +41,17 @@
     "Specify the IAM role for permission to access the dataset in S3 and SageMaker functionality."
```
iaroslav-ai, line 41: <p>Could you add a comment as a text above the cell? Right now it states 'specify the IAM role ... ` and then the cell with installation of scikit-image, which is a bit unexpected. IMO would look better like this: </p><p> </p><p> Specify IAM role</p><p><code>the iam role cell</code>  </p><p> </p><p>install scikit-image package </p><p><code>installation cell</code></p><p> </p><p>... </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/awslabs/amazon-sagemaker-examples/pull/1166/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

PR #1169, opened 2020-04-27
Update RL Ray launcher and accompanying example

annaluo676:
*Description of changes:*
1. Fix Ray 0.8.2 API breaking changes for cluster joining;
2. Update policy export logic for TensorFlow serving;
2. Avoid saving experiment config through `json.dump()` in case users have callbacks / customized model. Instead we use `params.json` generated by RLlib;
3. Update the roboschool example notebook to use the latest container.

*Tests:*
Changes tested on Roboschool example (single instance for deployment, multiple instances for homogenous/heterogeneous scaling) and BattleSnake experiments.

### DIFF
--- a/reinforcement_learning/common/sagemaker_rl/ray_launcher.py
+++ b/reinforcement_learning/common/sagemaker_rl/ray_launcher.py
@@ -139,7 +139,10 @@ class SageMakerRayLauncher(object):
             print("Waiting for %s worker nodes to join!" % (len(all_wokers_host_names)))
             self.sage_cluster_communicator.wait_for_signals(all_wokers_host_names)
             print("All worker nodes have joined the cluster. Now training...")
-            config = {"redis_address": "%s:6379" % master_ip}
+            if ray.__version__ >= "0.8.2":
+                config = {"address": "%s:6379" % master_ip}
+            else:
+                config = {"redis_address": "%s:6379" % master_ip}
         else:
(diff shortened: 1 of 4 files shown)


reinforcement_learning/common/sagemaker_rl/ray_launcher.py
```diff
@@ -206,8 +213,16 @@ def copy_checkpoints_to_model_output(self):
...
-            json.dump(config, f, indent=2)
+        config_found = False
+        for root, directories, filenames in os.walk(INTERMEDIATE_DIR):
+            if config_found:
+                break
+            else:
+                for filename in filenames:
+                    if filename == "params.json":
+                        source = os.path.join(root, filename)
+                        config_found = True
```
yijiezh, line 224: Why not break after `config_found` but in a different branch?

yijiezh: The original code simply dump the provided config into a JSON file. Here we simply ignore the provided config?

annaluo676: > Why not break after `config_found` but in a different branch?

This is to break the outer `for` loop `for root, directories, filenames in os.walk(INTERMEDIATE_DIR):`

>The original code simply dump the provided config into a JSON file. Here we simply ignore the provided config?

Yes. The original code breaks if any part of the config is not JSON serializable (e.g., customer policy). We use the config that RLlib generates instead.

Issue #746, opened 2019-05-23
from misc import get_execution_role, wait_for_s3_object


saunak14:
The above-mentioned `wait_for_s3_object` import statement in any of the reinforcement learning (rl) python notebook examples do not work. Please help.

yijiezz:
Hi @saunak14,

Can you elaborate a little bit on how it failed?
Does the line `from misc import get_execution_role, wait_for_s3_object` simply failed?

chester755:
Hi,

Yeah, I have a similar problem trying to run that on my local machine (as I want to see how that performance).
Below is the error message I get:
---------------------------------------------------------------------------
ModuleNotFoundError                       Traceback (most recent call last)
<ipython-input-2-8094e9abd582> in <module>
     10 from time import gmtime, strftime
     11 sys.path.append("common")
---> 12 from misc import get_execution_role, wait_for_s3_object
     13 from sagemaker.rl import RLEstimator, RLToolkit, RLFramework

ModuleNotFoundError: No module named 'misc'

From the research, it seems like the misc package is the AWS package.
Could you please help me with that?

sidd1809:
Hi @chester755 ,

Can you provide details on how you are using the example notebook?

'misc' and other modules are present in common/  and we append common/ to sys.path before importing any module. We moved this common module one level up, and symlink-ed  the common directory in each example notebooks. If you are copying the notebook folder else were, you might want to use -L flag with the cp  command as explained here: https://unix.stackexchange.com/questions/94714/cp-l-vs-cp-h

chester755:
Hi @sidd1809 

Thanks for your reply.
Sure I am basically git clone the repository into my windows machine, where I have Anaconda, Jupyter notebook installed,
After starting the Jupyter notebook and navigate to the example I want to try locally, I have started to run the example (rl_traveling_salesman_vehicle_routing_coach).
Two things are failed at the moment, one of them is can't find module sagemaker, the other one is the 'misc' module which I think I can bypass now.
The error message whenever I tried to install sagemaker through conda pip are as follows:

Installing collected packages: jmespath, botocore, s3transfer, boto3, protobuf, protobuf3-to-dict, PyYAML, idna, requests, cached-property, texttable, docopt, websocket-client, dockerpty, pypiwin32, pynacl, bcrypt, paramiko, docker, docker-compose, sagemaker
  Found existing installation: PyYAML 5.1.2
[...]

PR #1176, opened 2020-04-29
Adding reinforcement learning example from popular RL textbook

benjaminbluhm:
*Issue #, if available:*

*Description of changes:*

Adding a reinforcement learning example for stock trading taken from chapter 8 of textbook [Deep Reinforcement Learning Hands-On](https://github.com/PacktPublishing/Deep-Reinforcement-Learning-Hands-On)  by Maxim Lapan. 


benjaminbluhm:
Hi @hsl89 

I get the following build error:

```
     10 from time import gmtime, strftime
     11 sys.path.append("common")
---> 12 from misc import get_execution_role, wait_for_s3_object
     13 from sagemaker.rl import RLEstimator, RLToolkit, RLFramework

ModuleNotFoundError: No module named 'misc'
```
The issue is also discussed [here](https://github.com/aws/amazon-sagemaker-examples/issues/746). 

'misc' and other modules are present in common/ which is appended to sys.path before importing any module. I symlink-ed the common directory in the notebook directory the same way as is done for other RL notebooks in the repo. On my local notebook the import of the module works fine.

Any advise on how to fix this problem would be highly appreciated.

hsl89:
Hello @benjaminbluhm, I had the similar problem with other notebooks as well. At this point, it seems the CI cannot pick up the symbolic links (it is probably safe to assume CI cannot pick up anything outside the directory of the notebook). Hopefully we will have a better CI in the future. In the meantime, I think the best workaround is to simply test the notebook by hand and ignore the CI build error. When merging the PR, we should add a comment about why we need to "squash and merge" for future reference.

benjaminbluhm:
Hi **@hsl89** ,

Your suggestion for temporary workaround via manual testing makes sense to me. 

Thanks for your quick feedback and clarification!

PR #1205, opened 2020-05-07
Fix ECR URI suffix in Feature Transformation notebook

nigenda-amazon:
*Issue #, if available:*
Pushing ECR image failed in regions other than cn-north-1

*Description of changes:*
* Fix URL suffix when in commercial so that it's not a tuple
* Only set CN URI suffix when in cn-north-1 or cn-northwest-1


nigenda-amazon:
Tested by running in both a CMH and a ZHY notebook instance. Works.

PR #1212, opened 2020-05-12
Optimization to run the agents reward function at 15 steps per second

sunil19m:
This is to address the issue that the steps executed did not match the camera FPS. With this change if the sensor is running at 15FPS, the step function is called 15times.


ReHoss:
Hello, the notebook is broken when using presets files to custom the RL Coach algorithm (try for exemple a custom DQNAgent see https://github.com/aws-samples/aws-deepracer-workshops/tree/master/Advanced%20workshops/AI%20Driving%20Olympics%202019/challenge_train_DQN by @sahikagenc. Old simapp version.

Even when by-passing presets files and directly modifying the sagemaker_graph_manager.py, DQN does not work (problems when loading files).

Do you plan to remove presets files from the framwork and only keep the ClippedPPOAgent available ? I really need to switch from ClippedPPO to other algorithm for academic research purposes. I could help you.


Thank you for your consideration,

Kind regards,

Issue #1218, opened 2020-05-15
FutureWarning: Creating a DatetimeIndex by passing range endpoints is deprecated.  Use `pandas.date_range` instead.


gsamaras:
Using the latest code from the master branch ([7262f9e](https://github.com/awslabs/amazon-sagemaker-examples/commit/7262f9eca92a612423dba16595161c2a473b771e)), I received this warning:

`FutureWarning: Creating a DatetimeIndex by passing range endpoints is deprecated.  Use pandas.date_range instead.`

in the [DeepAR introduction notebook](https://github.com/awslabs/amazon-sagemaker-examples/blob/master/introduction_to_amazon_algorithms/deepar_synthetic/deepar_synthetic.ipynb), when executing the code in AWS Sagemaker Notebook (Jupyter lab).

gsamaras:
After this merge, the FutureWarnings disappeared, issue can be resolved.

PR #1219, opened 2020-05-15
Resolved all FutureWarnings of Panda in Deepar Introduction Notebook

gsamaras:
*Issue #, if available:*
[#1218](https://github.com/awslabs/amazon-sagemaker-examples/issues/1218)

*Description of changes:*
[The Deepar Introduction Notebook](https://github.com/awslabs/amazon-sagemaker-examples/blob/master/introduction_to_amazon_algorithms/deepar_synthetic/deepar_synthetic.ipynb) emitted three Pandas Future Warnings:

    FutureWarning: Creating a DatetimeIndex by passing range endpoints is deprecated. Use pandas.date_range instead.

twice, and:

    FutureWarning: Addition/subtraction of integers and integer-arrays to Timestamp is deprecated, will be removed in a future version.  Instead of adding/subtracting `n`, use `n * self.freq`

Since we insist on the highest standards, I took the liberty to resolve them, by replacing `pd.DatetimeIndex()` with `pd.date_range()`

and also

use `pd.Timedelta(1, unit=self.freq)` to increment the Timestamp, instead of just adding 1. Note that I used the frequency to dynamically select the unit (instead of statically passing the `hours=1` parameter), since in the top cells, the user defines `freq` as `'H'`, but if somebody changed that locally, then that change would not be reflected in the prediction code (which applies that increment).

I checked all the Notebook's cells after my changes, and everything worked as before (except from the warnings obviously), results were the same.

### DIFF
--- a/introduction_to_amazon_algorithms/deepar_synthetic/deepar_synthetic.ipynb
+++ b/introduction_to_amazon_algorithms/deepar_synthetic/deepar_synthetic.ipynb
@@ -202,7 +202,7 @@
     "    source = level + seas_amplitude*np.sin(time_ticks*(2*np.pi)/period)\n",
     "    noise = sig*np.random.randn(data_length)\n",
     "    data = source + noise\n",
-    "    index = pd.DatetimeIndex(start=t0, freq=freq, periods=data_length)\n",
+    "    index = pd.date_range(start=t0, freq=freq, periods=data_length)\n",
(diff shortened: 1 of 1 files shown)


fallingstar3107:
Hi, 

I still got the error "TypeError: __new__() got an unexpected keyword argument 'start' " even after changed pd.DatetimeIndex() to pd.date_range():
image

PR #1221, opened 2020-05-17
Port "R bring your own" notebook to R

athewsey:
*Issue #, if available:* None yet

*Description of changes:*

It seems strange to me that the notebook for the `advanced_functionality/r_bring_your_own` example is in Python:

- There are already loads of examples of Python notebooks elsewhere
- SageMaker Notebook Instances today include an R kernel by default (although not in Studio, I think)
- If somebody's going against the vast majority of samples and documentation out there by choosing R over Python, they probably are committed to not wanting to use Python!

In this fork I translated the notebook to R as faithfully as I could, with a couple of notable modifications:

1. Updated Docker base to ubuntu:20.04 (also an LTS release) from 16.04, because I was getting install errors for `mda` library.
2. Factored the container build & publish shell commands out into a script, because IRKernel doesn't support the `%%sh` or `!` magics like IPython does - so it was the only practical way to run the shell scripts from the notebook (It's still not ideal, because the output only comes through after the script succeeds or fails)
3. Upload data with sagemaker SDK's `upload_data()` function rather than the boto3 resource, as it's a little less verbose.
4. Tweaked down the suggested notebook instance type, because it works just fine on a t3.medium.

There are several more significant changes I think could also be useful for the example in future, too:
- Pinning versions of R and the installed packages, for better repeatability & preventing the kind of errors I was seeing with Ubuntu 16
- Improve the R package install so the Docker build actually errors on failed/missing packages, rather than warning & continuing
- Refactor to use the high-level Estimator API rather than the boto3 APIs
- Extend to work more like a framework container with "bring-your-own-script" functionality

...would be good to get opinions on whether those are of interest.

One important disclaimer: I'm newer to R than Python, so I'm very ready for some criticism of the code! 😂


athewsey:
^ I think this failing CI check might just be that it doesn't know how to process R notebooks at all? Since it's a syntax error on the first non-Python-like line of the notebook and that cell definitely runs for me!

paul-cb (approved):
Approved.

PR #1227, opened 2020-05-19
Add xgboost_abalone and sklearn_iris ipynb's to sagemaker_studio

edmrossi:
*Description of changes:*
Adding xgboost & sklearn examples to the sagemaker studio example set.  These notebooks have been run & verified using the Python 3 (Data Science) kernel


edmrossi:
Updated with cleared output

aws-patlin:
Sorry for the thrash, but do we want to update these for XGBoost 1.0-1? @iyerr3

iyerr3:
Yes, I think it's best if we update to `1.0-1`.

PR #1269, opened 2020-06-10
changed instance type so it can run with default

ngluna:
- Changed instance type from ml.p2.xlarge to ml.p3.2xlarge. Re: default quota exceeded. ml.p2.xlarge has a limit of 1 instance, ml.p3.2xlarge has a limit of 2

- Added link to instance pricing and estimate training time so customers get a sense of how much the instances they are spinning up cost.

- Set "wait" parameters in "estimator.fit()"  method to "True". Re: If set to "False" the notebook will crash if a user does a "Run All Cells". If set to"False" notebooks returns immediately, which allows the user to then execute next training example (e.g. Pipe mode ex. to Horovod ex.) thus returning a "ResourceLimitExceeded" error.

- Removed code to view TensorBoard code and text mentioning tensorboard. Re:  Doesn't work. Unable to configure a local host within SageMaker Studio.

- Removed 'wait' parameter in estimator.fit() because by default it is True.


aws_sagemaker_studio/frameworks/keras_pipe_mode_horovod/keras_pipe_mode_horovod_cifar10.ipynb
```diff
@@ -602,9 +609,9 @@
  "metadata": {
   "instance_type": "ml.t3.medium",
   "kernelspec": {
-   "display_name": "Python 3 (TensorFlow CPU Optimized)",
+   "display_name": "Python 3",
```
aaronmarkham, line 612: Not sure how much this matters in this example, but I think generally, for Keras, you'd want the TF Optimized kernel.

aws_sagemaker_studio/frameworks/keras_pipe_mode_horovod/keras_pipe_mode_horovod_cifar10.ipynb
```diff
@@ -457,7 +464,7 @@
    },
    "outputs": [],
    "source": [
-    "predictor = estimator.deploy(initial_instance_count=1, instance_type='ml.m4.xlarge')"
+    "predictor = estimator.deploy(initial_instance_count=1, instance_type='ml.p3.2xlarge')"
```
aaronmarkham, line 467: I think you can leave this one as m4. Did it not work?

aws_sagemaker_studio/frameworks/keras_pipe_mode_horovod/keras_pipe_mode_horovod_cifar10.ipynb
```diff
@@ -273,26 +280,19 @@
...
-   ]
-  },
   {
    "cell_type": "code",
    "execution_count": null,
    "metadata": {},
    "outputs": [],
    "source": [
-    "pipe_mode_estimator.fit(inputs, wait=False)"
+    "pipe_mode_estimator.fit(inputs, wait=True)"
```
laurenyu, line 295: `wait=True` is the default, so you can just remove it

aws_sagemaker_studio/frameworks/keras_pipe_mode_horovod/keras_pipe_mode_horovod_cifar10.ipynb
```diff
@@ -597,14 +552,21 @@
...
     "predictor.delete_endpoint()"
    ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": []
   }
```
laurenyu, line 562: remove the extra cell

aws_sagemaker_studio/frameworks/keras_pipe_mode_horovod/keras_pipe_mode_horovod_cifar10.ipynb
```diff
@@ -380,60 +373,18 @@
...
-    "After running that command, we can access TensorBoard locally at http://localhost:6006.\n",
-    "\n",
-    "Based on the TensorBoard metrics, we can see that:\n",
-    "1. All jobs run for 10 epochs (0 - 9).\n",
-    "1. Both File Mode and Pipe Mode run for ~1 minute - Pipe Mode doesn't affect training performance.\n",
-    "1. Distributed training runs for only 45 seconds.\n",
-    "1. All of the training jobs resulted in similar validation accuracy.\n",
-    "\n",
-    "This example uses a relatively small dataset (179 MB). For larger datasets, Pipe Mode can significantly reduce training time because it does not copy the entire dataset into local memory."
+    "dist_estimator.fit(inputs, wait=True)"
```
laurenyu, line 387: you can also remove "wait=True" here

aws_sagemaker_studio/frameworks/keras_pipe_mode_horovod/keras_pipe_mode_horovod_cifar10.ipynb
```diff
@@ -444,7 +395,11 @@
     "\n",
     "After we train our model, we can deploy it to a SageMaker Endpoint, which serves prediction requests in real-time. To do so, we simply call `deploy()` on our estimator, passing in the desired number of instances and instance type for the endpoint.\n",
     "\n",
-    "Because we're using TensorFlow Serving for deployment, our training script saves the model in TensorFlow's SavedModel format. For more details, see [this blog post on deploying Keras and TF models in SageMaker](https://aws.amazon.com/blogs/machine-learning/deploy-trained-keras-or-tensorflow-models-using-amazon-sagemaker)."
+    "Because we're using TensorFlow Serving for deployment, our training script saves the model in TensorFlow's SavedModel format.  \n",
```
laurenyu, line 398: [nitpick]
```suggestion
    "Because we're using TensorFlow Serving for deployment, our training script saves the model in TensorFlow's SavedModel format.\n",
```

aws_sagemaker_studio/frameworks/keras_pipe_mode_horovod/keras_pipe_mode_horovod_cifar10.ipynb
laurenyu: The reason I chose the TF kernel over the data science one was because the code for generating the training data requires `tensorflow` to be installed in the environment. (though I don't think the display name really matters all that much)

PR #1270, opened 2020-06-10
Upgrade notebook with latest XGB image

balajitummala:
*Description of changes:*
Updated the notebook with the latest xgboost image that supports Multi model Endpoint


balajitummala:
@eslesar-aws Could you please review ? 
As we released XGBoost multi model endpoint support to the latest image, we don't need to build from that branch anymore.

PR #1319, opened 2020-07-10
as_matrix is deprecated and breaks current versions

javier:
as_matrix is deprecated and breaks current versions, including the default notebooks created at sagemaker

*Issue #, if available:*

*Description of changes:*


paul-cb:
Thanks @javier for these changes!

I tested this end-to-end, it breaks because sagemaker is not imported. Could you fix this and ensure it runs end-to-end and also specify what Kernel and version of Sagemaker one should be using? 

```
---------------------------------------------------------------------------
NameError                                 Traceback (most recent call last)
<ipython-input-1-276df1ea7e06> in <module>
      6 role = get_execution_role()
      7 
----> 8 bucket = sagemaker.Session().default_bucket()# enter your s3 bucket where you will copy data and model artifacts
      9 prefix = 'sagemaker/DEMO-breast-cancer-prediction' # place to upload training files within the bucket

NameError: name 'sagemaker' is not defined
```

paul-cb:
Thanks @javier! The notebook runs end-to-end but requires a customer to specify their bucket. One minor suggestion is to add `import sagemaker` and set`bucket = sagemaker.Session().default_bucket()`. Customers often specify their bucket in this way and when they do in this case, it raises an import error because sagemaker wasn't imported. Could you make these changes?

paul-cb (approved):
Approved.

PR #1335, opened 2020-07-18
Joblib deprecated

x18183468:
Hello,
Joblib is removed in the new version of SKLearn due to which from sklearn.externals import joblib does not work, import joblib works. Local execution fails but in EC2 it executes well.

Thanks

*Issue #, from sklearn.externals import joblib has been removed. import  joblib works. 

*Description of changes:* Although local execution fails but in EC2 of ML, execution succeeds.


aaronmarkham:
Another PR also fixed this issue and that one has been merged. Thanks for the contribution though!!

PR #1622, opened 2020-10-14
adding an example for Linear Learner regression use case

moagaber:
… dataset and input csv format

*Issue #, if available:*

*Description of changes:*
This is a Jupyter notebook to explain how to use Amazon SageMaker Linear Learner algorithm to solve regression use case using abalone data-set. It also demonstrate how we can use the algorithm with input files of csv format.


paul-cb (approved):
Tested this notebook. It runs

PR #1634, opened 2020-10-20
Add SDK version rollback code.

Laren-AWS:
Added code to store the original SDK version to a file before the SDK is downgraded to 1.72, and to roll the version back to the original SDK version after the notebook completes.


aaronmarkham (approved):
LGTM

PR #1651, opened 2020-10-23
Separate MXNet and Tensorflow code for coach-launcher

goelakash:
*Issue #, if available:*

*Description of changes:*
RL Coach notebooks use MXNet in the backend, but the code also imports tensorflow. This causes the notebooks to error out during training, as the docker image for training doesn't have tensorflow. This change will fix the behaviour so that only notebooks that need tensorflow will import it.


goelakash:
Closing this in favor of #1652

PR #1669, opened 2020-10-29
Fixing RL notebooks 

goelakash:
*Issue #, if available:*

*Description of changes:*
(see commits)


goelakash:
FYI the codebuild test fails for notebooks that are already broken and not covered by this PR:

reinforcement_learning/common/sagemaker_rl/ray_launcher.py
```diff
@@ -123,7 +123,7 @@ def get_all_host_names(self):
 
     def ray_init_config(self):
         num_workers = max(self.num_cpus, 3)
-        config = {"num_cpus": num_workers, "num_gpus": self.num_gpus, "webui_host": '127.0.0.1'}
```
annaluo676, line 126: Any specific reason to remove `webui_host`? iirc this will suppress the misleading ray dashboard error.

reinforcement_learning/common/sagemaker_rl/tf_serving_utils.py
```diff
@@ -1,9 +1,14 @@
...
-from ray.rllib.utils.framework import try_import_tf
+import tensorflow as tf
 
-tf = try_import_tf()
+def import_tf1():
+    if "2." in tf.__version__[:2]:
+        return tf.compat.v1
+    return tf
+
+tf = import_tf1()
```
annaluo676, line 11: how is this different from ray's native `try_import_tf`? That method also tries to switch off v2_behavior.

reinforcement_learning/common/sagemaker_rl/tf_serving_utils.py
```diff
@@ -30,7 +35,7 @@ def export_tf_serving(agent, output_dir):
         input_signature["observations"] = tf.saved_model.utils.build_tensor_info(policy.observations)
 
         output_signature = {}
-        output_signature["actions"] = tf.saved_model.utils.build_tensor_info(policy.sampler)
+        output_signature["actions"] = tf.saved_model.utils.build_tensor_info(policy._sampler)
```
annaluo676, line 38: Good catch! This seems to depend on the TF version as well. Would it break anything if `policy` is using tf2?

reinforcement_learning/common/sagemaker_rl/ray_launcher.py
goelakash: This is to make sure that the examples run using Ray 0.6.5, which is the one supported by SDK v1.72. `webui_host` seems to have been introduced to the init function in later versions of Ray.

reinforcement_learning/common/sagemaker_rl/tf_serving_utils.py
goelakash: That method is only available in later versions of Ray, and not in 0.6.5.

yijiezh: We might need some more comments here to explain the context. I think we just switching from `tf` to `try_import_tf` in this commit: https://github.com/aws/amazon-sagemaker-examples/commit/44d0ffa9d2e90b90fa64fbc8e987012fe4b02250#diff-68432ce762a5889a6722a6ff6184996fd732f6cf8a9d6ca00b0d449f21ef5522

reinforcement_learning/common/sagemaker_rl/tf_serving_utils.py
goelakash: So I ran the network_node_compression example after this change (and another, by replacing `logits` with `_action_prob`), and it worked fine.

goelakash:
Closing this PR, since these changes will not be needed with the latest planned updates (SDK 2.0 and Ray 1.0).

reinforcement_learning/common/sagemaker_rl/tf_serving_utils.py
leijin168: 1) I thought we are targeting ray 1.0 soon/going forward. so try_import_tf is safe here?

reinforcement_learning/rl_network_compression_ray_custom/src/train-ray.py
```diff
@@ -39,15 +39,15 @@ def get_experiment_config(self):
             "checkpoint_freq" : 1,
             "config": {
               "num_workers": max(self.num_total_gpus-1, 1),
-              "use_gpu_for_workers": True,
+              "num_gpus_per_worker": 1,
```
leijin168, line 42: I assume these are the config name changes from 0.5.3 to 0.6.5?

PR #1761, opened 2020-11-18
fix unzip package missing, fixed the container version contract

vikeshpandey:
*Issue #, if available:*
the unzip utility installation steps fails on studio. xgboost container expects a version

*Description of changes:*
have replaced the unzip package code with simpler python unzip module code which works on studio or on any other env. the earlier code was dependent on package manager which is linux flavour specific and the cell breaks in studio. along with that, the get_image_uri contract expects a version to be supplied.

along with it, the code is also upgraded to SDK v2


aaronmarkham (changes requested):
Check the build logs... 
Still needs "<your_s3_bucket_name_here>" replaced with default bucket.

aaronmarkham (approved):

PR #1790, opened 2020-11-24
Update  with sagemakerV2 functions

preshen-goobiah:
*Issue #, if available:*

Notebook fails on estimator requiring "image_uri".
Warning thrown for "sagemaker.amazon_estimator.get_image_uri()"

*Description of changes:*
Update Estimators to use "image_uri"
Get image uri with "sagemaker.image_uris.retrieve()"


seanpmorgan (changes requested):
@preshen-goobiah Thanks for the PR! The CI is failing since the first cell installs s3fs and downgrades sagemaker SDK. s3fs is already available in the kernel though so could you please delete the first cell and then this should work.

PR #1851, opened 2020-12-10
Fix py_version, remove dllogger, workaround checkpoint issue in SM

anirudh2290:
*Description of changes:*
- Change py_version to py36
- Remove dllogger usage
- Add sync of checkpoints to s3 bucket.

### DIFF
--- a/training/distributed_training/pytorch/model_parallel/bert/smp_bert_tutorial.ipynb
+++ b/training/distributed_training/pytorch/model_parallel/bert/smp_bert_tutorial.ipynb
@@ -53,9 +53,33 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "## Amazon SageMaker Initialization\n",
-    "\n",
-    "Initialize the notebook instance. Get the AWS Region, SageMaker execution role Amazon Resource Name (ARN). "
+    "## Amazon SageMaker Initialization"
+   ]
+  },
+  {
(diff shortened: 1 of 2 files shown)


TEChopra1000 (approved):

training/distributed_training/pytorch/model_parallel/bert/smp_bert_tutorial.ipynb
```diff
@@ -53,9 +53,33 @@
    "cell_type": "markdown",
```
aaronmarkham, line 53: <p>Do you need to upgrade for some reason? This can break the customer experience for subsequent notebooks, so if you must upgrade, you should rollback at the end of the notebook.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/aws/amazon-sagemaker-examples/pull/1851/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='4'/>

anirudh2290: The upgrade is required to make the Sagemaker SDK use the latest DLC, which comes with SMP. I had a rollback step in here (based on your example) which I removed because the python upgrades require kernel restart and the "original_version" is wrong after the restart. Is there a way to workaround the kernel restart ?

PR #1889, opened 2020-12-30
Set drop_last=True for DataLoader

anirudh2290:
*Issue #, if available:*

*Description of changes:* Set drop_last=True for DataLoader. This has become a common pitfall (where the batch size of last batch is not divisible by number of microbatches) and better to set it to True by default. Downside is if the total number of samples is smaller than batch size in which case there is no data, and may cause failures (which I am guessing is less likely compared to customers running into "batch size not divisible by num_microbatches" issue).

@hsl89


hsl89 (approved):
awaiting approval for using small sample of data for CI. Changes tested end-to-end manually

PR #1893, opened 2020-12-31
Updated linear_time_series_forecast.ipynb to support SageMaker API v2.x

dinhhn:


aaronmarkham (changes requested):
It's not passing CI because of this:
Invalid bucket name "<your_s3_bucket_name_here>"
Can you change that to use the default bucket?

dinhhn:
The bucket parameter appeared to be hardcoded that way for quite some time, so I didn't change it in case there was a reason for it.  However, since you are giving the green light to change it, it makes sense to set it to the default bucket as you suggested.  I have updated the bucket parameter to use the default bucket.  Please review.

aaronmarkham (approved):

PR #1924, opened 2021-01-08
update names of renamed serializers/deserializers in v2

seanlee10:
*Issue #, if available:*

*Description of changes:*
While following thru the existing notebook, I realized it's using sagemaker v1 references. I renamed them to align with the new names in v2 and was able to finish the example with the fix.


aaronmarkham (changes requested):
Can you also fix the "your bucketname here" error that CI is getting? You can just use the default bucket instead and it should pass CI with that...

seanlee10:
> 
> 
> Can you also fix the "your bucketname here" error that CI is getting? You can just use the default bucket instead and it should pass CI with that...

Thanks for the guidance. I updated my PR

aaronmarkham (approved):

PR #1968, opened 2021-02-02
Removed pip installation; Re: SM SDK is now by default v2.

ngluna:
*Issue #, if available:*
Removed pip install command to upgrade SageMaker SDK to v>=2. Re: v2 is now the default.

*Description of changes:*
Removed cells w/ pip command.


ngluna:
What kernel does CI use? 
The default when you launch this NI instance is conda_pytorch_p36 . Torch is included in this container. 
Torch is not included in conda_python3 .

ngluna:
The kernel used is the same as the data science kernel, hence, why  CI is breaking. [It needs to use the one specified by default in the notebook config.] I am overwriting and merging.

Issue #1958, opened 2021-01-27
failed to load image in groundtruth streaming label job


huckw:
I am using the ground_truth_create_streaming_labeling_job.ipynb and I am able to create the labeling job and publish some images to it.  But the workers (that includes me) get a "failed to load image" message and there seems to be nothing they can do to move forward.  When I stop the labeling job I can see the unlabled images in the output folder so I know my image url is legit.  What might I be doing wrong?

TEChopra1000:
@huckw did you use the default bounding box template in the notebook, or did you provide your own worker task template when you created the labeling job?

huckw:
I am using the SemanticSegmentation job instead of BoundingBox

huckw:
I will also mention that if I manually create a non-streaming labeling job with the console and use an input manifest that uses the same references to the same images then it works fine.  I made sure that my CORS for the bucket was set correctly.  Here is the template I am using...
```
    <script src="https://assets.crowd.aws/crowd-html-elements.js"></script>
    <crowd-form>
      <crowd-semantic-segmentation
        name="crowd-semantic-segmentation"
        src="{{ task.input.taskObject | grant_read_access }}"
        header="Please use the polygon tool to select the 4 corners of the ID.  If one or more corners fall off of the image select 'nothing to label'"
        labels="{{ task.input.labels | to_json | escape }}"
      >
        <full-instructions header="Segmentation instructions">
          <ol><li><strong>Read</strong> the task carefully and inspect the image.</li>
          <li><strong>Read</strong> the options and review the examples provided to understand more about the labels.</li>
          <li><strong>Choose</strong> the appropriate label that best suits an object and paint that object using the tools provided.</li></ol>
        </full-instructions>
        <short-instructions>
          <h2><span style="color: rgb(0, 138, 0);">Good example</span></h2>
          <p>Use polygon tool to select the four corners that best outline the ID</p>
          <p><br></p><h2><span style="color: rgb(230, 0, 0);">Bad example</span></h2>
          <p>Do not crop out any of the ID</p>
        </short-instructions>
      </crowd-semantic-segmentation>
    </crowd-form>
```

TEChopra1000:
I see, when you inserted this template into the notebook, did you add extra curly brackets `{}` around the Liquid elements?

For example, the bounding box example has an extra pair of `{}` for every pair of `{}` in the template. So given that the liquid elements have two pairs normally, there are four pairs here: 

```
[...]

PR #1981, opened 2021-02-08
documentation: adding a note about using liquid elements template

TEChopra1000:
*Issue #, if available:*
#1958 

*Description of changes:*
Adding note to tell users to add extra pair of `{}` around liquid elements if uploading one's own template. 


paul-cb (approved):
Looks good, approved.

Issue #1966, opened 2021-02-02
xgboost_costomer_churn notebook example and its variations fail due to the data download "Connection timed out"


mchoi8739:
The [xgboost_customer_chrun notebook](https://github.com/aws/amazon-sagemaker-examples/blob/master/introduction_to_applying_machine_learning/xgboost_customer_churn/xgboost_customer_churn.ipynb) and it's variations are not working due to the dataset download failure.

error message:
```
Connecting to dataminingconsultant.com (dataminingconsultant.com)|160.153.91.162|:80... failed: Connection timed out.
```

Might have to host the customer churn data through the SageMaker public S3 dataset bucket.

seanpmorgan:
Duplicate of https://github.com/aws/amazon-sagemaker-examples/issues/1950 . Does this dataset license allow us to redistribute?

hsl89:
I cannot find any legal restrictions on their website http://dataminingconsultant.com/

mchoi8739:
I explored the website as well, and the customer churn dataset is moved around and named differently too.

gategill:
It says their account has been suspended

gategill:
I'm new here and I'm trying to do this notebook too.

I found the data on Kaggle, the only difference is that column names are different and a few other caveats, like the number of rows.

To use the data in the notebook, I uploaded the "train.csv" file on GitHub and copied the raw link. It should look something like: 
`!wget https://raw.githubusercontent.com/[your_github_username]/amazon-sagemaker-examples/master/introduction_to_applying_machine_learning/xgboost_customer_churn/train.csv
`

There's no need to unzip it as it's already a `.csv` file. Just use pandas to read it in `churn = pd.read_csv("./train.csv")`

I'm just starting the notebook example and it's been working fine. Hope this helps you.



[Data on Kaggle](https://www.kaggle.com/c/customer-churn-prediction-2020/data)

hsl89:
hey @gategill, the main issue here is can we legally use the data set (from kaggle or the original hosting site)

seanpmorgan:
Just to update, working on synthetic dataset. Should be ready later today

seanpmorgan:
New dataset and download commands have merged. Closing, and thanks for reporting!

PR #1999, opened 2021-02-16
Update dataset path exaplaining customer churn autopilot

seanpmorgan:
Part of https://github.com/aws/amazon-sagemaker-examples/issues/1966


seanpmorgan:
Due to the way this notebook is structured this will not pass CI without a prior automl job. IMO it should be removed or restructured to include the job, but either way this PR shouldn't be gated

PR #2017, opened 2021-02-18
use_cases/customer_churn example

aarsanjani:
*Issue #, if available:*

*Description of changes:* added use_cases/customer_churn example for music use-case


mchoi8739:
Can you fix this PR to not include e2e notebooks that are already published?

aarsanjani:
I have submitted a subsequent one which is more valid

PR #2053, opened 2021-03-05
Updated to AutoGluon Tabular example to autogluon==0.1.0

araitats:
*Issue #, if available:*
AutoGluon was updated to ver. 0.1.0. The update was reflected to the sagemaker example. 

*Description of changes:*
Tabularprediction is no longer used. It is called TabularPredictor as of version 0.1.0 


advanced_functionality/autogluon-tabular/container-training/train.py
```diff
@@ -128,15 +136,16 @@ def train(args):
         pickle.dump(column_dict, f)
     
     # Train models
-    predictor = task.fit(
-        train_data=train_data,
-        output_directory=args.model_dir,
-        **args.fit_args,
-    )
+
+    args.fit_args.pop('label', None)
```
Innixma, line 140: perhaps also let users specify `eval_metric, sample_weight, weight_evaluation, learner_kwargs`? These are specified in init, goes in same place as label. ([doc link](https://auto.gluon.ai/stable/_modules/autogluon/tabular/predictor/predictor.html#TabularPredictor))

Perhaps we can call them `args.init_args`?

Innixma: This also means `label` can be moved to `init_args`, or made as its own things similar to `args.model_dir`

Innixma: yes, you could also maybe just have model_dir be in init_args as `path` and remove `model_dir`, but unsure the reasoning behind current design.

araitats: `model_dir` is ENV VAR off of sagemaker. It is safe to do it this way.

araitats:
git reset --hard a2ee03c to see whether it passes sagemaker-bot or not.

Innixma (approved):

advanced_functionality/autogluon-tabular/AutoGluon_Tabular_SageMaker.ipynb
```diff
@@ -76,8 +76,8 @@
     "    )\n",
```
hsl89, line 76: <p>Build fails in this cell. Error is:</p><pre class="ql-syntax" spellcheck="false">s3://sagemaker-us-west-2-521695447989/autogluon-sagemaker-training-2021-03-11-04-17-28-887/output/model.tar.gz"
</pre><p> </p><p>my guess is the <code>model_uri</code>  you defined as <code>os.path.join(estimator.output_path, estimator._current_job_name, "output", "model.tar.gz")</code>  is not where the model is. Try to use <code>estimator.model_data</code>  directly. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/aws/amazon-sagemaker-examples/pull/2053/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='27'/>

araitats: Fixed. 
model_uri = estimator.model_data. See commit : bcf58a6

PR #2058, opened 2021-03-08
more robust way to download model artifact

hsl89:
*Issue #, if available:*

*Description of changes:*
deploy a tensorflow model from a public bucket 
fix https://github.com/aws/amazon-sagemaker-examples/pull/1848#issuecomment-780262411

### DIFF
--- a/training/distributed_training/tensorflow/data_parallel/mnist/infer_tensorflow.ipynb
+++ b/training/distributed_training/tensorflow/data_parallel/mnist/infer_tensorflow.ipynb
@@ -12,9 +12,31 @@
     "\n",
     "This notebook uses a stored model if it exists. If you recently ran a training example that use the `%store%` magic, it will be restored in the next cell.\n",
     "\n",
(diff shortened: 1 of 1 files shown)


training/distributed_training/tensorflow/data_parallel/mnist/infer_tensorflow.ipynb
```diff
@@ -72,62 +98,6 @@
     "predictor = model.deploy(initial_instance_count=1, instance_type='ml.m4.xlarge')"
    ]
   },
-  {
```
IvyBazan, line 75: Why has the endpoint test been removed?

hsl89: because so many notebooks already have the exactly the same test and I think it is better to make the content more modular. we can it by having one notebook elaborate on tensorflow model deployment and link other notebooks to it

PR #2070, opened 2021-03-10
Update sagemaker_featurestore_fraud_detection_python_sdk.ipynb

can-sun:
*Description of changes:*
Recently SageMaker FeatureStore released a new field called ResolvedOutputS3Uri, now we can directly get the destination of S3Uri where the offline data is stored. So previous string manipulation is no longer needed.


paul-cb (approved):
Looks good. Approved.

PR #2073, opened 2021-03-11
Create endpoint

hsl89:
*Issue #, if available:*

*Description of changes:*
added notebook for `CreateEndpoint`


sagemaker-fundamentals/create-endpoint/container/src/predictor.py
```diff
@@ -0,0 +1,42 @@
+# This is the file that implements a flask server to do inferences. It's the file that you will modify to adapt to your own inference logic.
```
paul-cb, line 1: This file implements a flask server for inference. You can modify the file to align with your own inference logic.

sagemaker-fundamentals/create-endpoint/container/src/predictor.py
```diff
@@ -0,0 +1,42 @@
...
+    """Determine if the container is working and healthy. 
+    In this sample container, we declare
+    it healthy if we can load the model successfully."""
+
+    status = 200
+    return flask.Response(response='\n', 
+        status=status, mimetype='application/json')
+
+@app.route('/invocations', methods=['POST'])
+def inference():
```
paul-cb, line 20: Does this method actually do inference? It seems like it only prints. If so, then the method name should be something slightly different.

sagemaker-fundamentals/create-endpoint/container/src/predictor.py
```diff
@@ -0,0 +1,42 @@
...
+    """Do an inference on incoming data.
```
paul-cb, line 21: "Perform" inference instead of "Do"

sagemaker-fundamentals/create-endpoint/container/src/serve
```diff
@@ -0,0 +1,69 @@
+#!/usr/bin/env python
+
+# This file implements the scoring service shell. You don't necessarily need to modify it for various
```
paul-cb, line 3: You don't necessarily need to modify it for various algorithms. --> You can modify it as need be for your algorithms.

sagemaker-fundamentals/create-endpoint/container/src/serve
```diff
@@ -0,0 +1,69 @@
...
+import signal
+import subprocess
+import sys
+
+cpu_count = multiprocessing.cpu_count()
+
+model_server_timeout = os.environ.get('MODEL_SERVER_TIMEOUT', 60)
+model_server_workers = int(os.environ.get('MODEL_SERVER_WORKERS', cpu_count))
+
+def sigterm_handler(nginx_pid, gunicorn_pid):
```
paul-cb, line 27: Add doc string.

sagemaker-fundamentals/create-endpoint/container/src/serve
hsl89: done

sagemaker-fundamentals/create-endpoint/container/src/predictor.py
hsl89: done

sagemaker-fundamentals/create-endpoint/container/src/predictor.py
hsl89: done

paul-cb (approved):
Looks good.

sagemaker-fundamentals/create-endpoint/container/src/predictor.py
hsl89: `inference` is what will be called when you invoke the endpoint with SageMaker runtime client. In real life an ML practitioner should write the inference logic under `def inference`. but here I just want to demonstrate that `def inference` is indeeded being called when we invoke the endpoint with sagemaker runtime. It prints out the request, indicating that it received the request

PR #2077, opened 2021-03-12
Maskrcnn updates to use SageMaker Script Mode

ajayvohra2005:
### Mask R-CNN container and notebook updates to use Script Mode

Up until now, the Mask RCNN containers contained the entry script invoked by SageMaker, and the entry script managed the distributed multi-machine MPI training. In this update, I have created additional containers, and notebooks that use SageMaker Script Mode. This simplifies the Containers, because they no longer contain the entry script. Now the entry script is outside the container, and is submitted to SageMaker Training Job through the TensorFlow Estimator. The old notebooks have been marked deprecated in the Readme file.


hsl89:
Hi @ajayvohra2005 since SageMaker does not support cross-region data access, when you ask users to interactively fill the name of their s3 bucket or file system id, it would be helpful to instruct them that the bucket should be in the same region as their vpc. And their sagemaker service role have access to the bucket. You can put those general instructions in the readme

ajayvohra2005:
> Hi @ajayvohra2005 since SageMaker does not support cross-region data access, when you ask users to interactively fill the name of their s3 bucket or file system id, it would be helpful to instruct them that the bucket should be in the same region as their vpc. And their sagemaker service role have access to the bucket. You can put those general instructions in the readme

I am adding code to the notebook  to validate that S3 bucket is located in same AWS region.

hsl89 (approved):

hsl89:
@ajayvohra2005 @ajaykarpur @aaronmarkham  If no one object, I will merge it tomorrow evening. Then we can move forward as discussed.

PR #2081, opened 2021-03-15
Use smp.amp.GradScaler for fp16 training in model-parallel PT BERT

karakusc:
`fp16` training currently does not work properly in the model parallel BERT training example. This PR switches to the built-in `smp.amp.GradScaler` API so that the example works with the flag `fp16: 1`. Also enables `fp16` by default.


hsl89:
@ajaykarpur, the build log seems to suggest sagemaker pysdk used in this PR and CI are different. Is CI always using the most up-to-date version pysdk?

zhaoqizqwang:
Closing this PR since it was created a long time ago and didn't pass all the checks. Feel free to open a new pull request instead.

PR #2082, opened 2021-03-16
Change ray 0.8.2 to 0.8.5.

tao2020:
*Issue #, if available:*
1. sagemaker-rl-ray-container:ray-0.8.2-tf will be deprecated and changed to 0.8.5.
2. Fix the no variable "job_name" issue.

*Description of changes:*
1. Change ray TF container from 0.8.2 to 0.8.5 in the Dockerfile. We have verified 0.8.5 is working correctly. 
Also add one sentence to the notebook to alert the customers the TrainingJobAnalytics will only work after the training is done.  
2. change `estimator.fit(wait=True)`, add `debugger_hook_config=False` to the RLEstimator 


sahikagenc (dismissed):
This updates the Unity ML notebook to Ray 0.8.5. It provides comment on the training metrocs.

PR #2166, opened 2021-04-27
fix: xgboost model using EI although not supported

georschi:
in this example the xgboost model was created for batch transform with an accelerator defined although this is not supported by the algorithm. 
It is not failing however, it is misleading. 
According to this: https://docs.aws.amazon.com/sagemaker/latest/dg/ei.html
as well as this: https://github.com/aws/sagemaker-python-sdk/blob/master/src/sagemaker/xgboost/model.py#L136 
we can deduce that ei accelerators are not supported by xgboost models

*Description of changes:*
deleted the addition of "accelerator_type" parameter for model definition


zhaoqizqwang:
Closing this PR since it was created a long time ago and didn't pass all the checks. Feel free to open a new pull request instead.

PR #2183, opened 2021-05-03
Update Mount Directory for FSx for Lustre

sayonkumarsaha:
*Issue #, if available:* NA

*Description of changes:*

The updated *directory_path requirement* for mounting data from S3 via FSx Lustre works with `/<mount-id>/<s3-prefix>`. 


ajayvohra2005:
I think this merge may break the notebook, because if one creates the FSx file-system from within the notebook, it creates a Scratch 1 FSx file-system, and the change in this pull-request is not needed. I think this merge will break the notebook for users creating the FSx file system from within the notebook.

sayonkumarsaha:
> I think this merge may break the notebook, because if one creates the FSx file-system from within the notebook, it creates a Scratch 1 FSx file-system, and the change in this pull-request is not needed. I think this merge will break the notebook for users creating the FSx file system from within the notebook.

The change I suggested is the `directory_path` requirement for S3 path in FSx in the data channel section of `mask-rcnn-fsx.ipynb` — Even if FSx file-system is created from outside, it currently doesn’t work with `/fsx/<s3-prefix>`  but requires `/<mount-id>/<s3-prefix>`. 

It’s a super useful notebook and this should help the customer to skip confusion since some of the FSX documentation [[blog](https://aws.amazon.com/blogs/machine-learning/speed-up-training-on-amazon-sagemaker-using-amazon-efs-or-amazon-fsx-for-lustre-file-systems/) | [doc](https://sagemaker.readthedocs.io/en/stable/overview.html#use-file-systems-as-training-inputs)] are also outdated — I’ve reached out to the Service team for the documentation.

ajayvohra2005:
*Even if FSx file-system is created from outside, it currently doesn’t work with /fsx/<s3-prefix> but requires /<mount-id>/<s3-prefix>*

Can you please confirm you first created a stack using the README and then you created the FSx from within the notebook cell? I tested it without this merge, and it continues to work without any error. I think the change above is only applicable to FSx SCRATCH-2 file-systems, not SCRATCH-1. That is my best understanding at the moment.

sayonkumarsaha:
You’re right that it works when you use the provided stack to create the FSx because the notebook creates the older FSx `SCRATCH-1` before the update was rolled out. In my case, the user already had a `SCRATCH-2` FSx created separately for loading data from S3 — and was using the notebook to run rest of the functionalities. Thus `/fsx/<s3-prefix>` didn’t work.

Based on the feedback, I also noticed that [this FSx blog](https://aws.amazon.com/blogs/machine-learning/speed-up-training-on-amazon-sagemaker-using-amazon-efs-or-amazon-fsx-for-lustre-file-systems/) has been updated which had similar issues with outdated deployment.

In that case, I can think of two options here:

- Revert back this PR merge. This is the valid an easy option — with the drawback being some customers refer to this as an example FSx + SM notebook (without running it end to end) for their own custom workload where they actually have the `SCRATCH_2` deployed FSx (it being updated with higher burst throughput capacity than `SCRATCH_1`). So could be confusing mount-names when someone doesn’t create the stack with the notebook.

- Other option is on top of my PR, update the LustreConfiguration in [cfn-fsx.yaml ](https://github.com/aws/amazon-sagemaker-examples/blob/master/advanced_functionality/distributed_tensorflow_mask_rcnn/cfn-fsx.yaml)with `DeploymentType: "SCRATCH_2”` as mentioned [here](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-fsx-filesystem.html), so that customers use the latest deployments.

What do you think?

PR #2192, opened 2021-05-04
Kernel reset

hsl89:
*Issue #, if available:*

*Description of changes:*
Set notebooks to use correct kernels 


aaronmarkham (changes requested):
Diffs are too big. Should just change the one line for the kernel for each file and nothing more.

PR #2641, opened 2021-05-29
Fix sentiment-analysis-tf-distributed-training-bringyourownscript.ipynb

hsl89:
Link to the notebook:
https://github.com/aws/amazon-sagemaker-examples/blob/master/sagemaker-debugger/tensorflow_nlp_sentiment_analysis/sentiment-analysis-tf-distributed-training-bringyourownscript.ipynb


mchoi8739 (changes requested):
Can you provide more information regarding your decision to put the top installation cell and forcing every library to be updated even though they exist?

As I mentioned before, it will be better to script the installation cell using the following sample:

```
import sys
import pip

def import_or_install(package):
    try:
        __import__(package)
    except ImportError:
        !{sys.executable} -m pip install {package}

required_packages=['smdebug', 'seaborn', 'plotly', 'opencv-python', 'shap', 'bokeh', 'imageio', 'sagemaker']

for package in required_packages:
    import_or_install(package)
```

jkroll-aws (approved):

PR #2717, opened 2021-06-15
Fix pytorch_smdataparallel_mnist_demo.ipynb

hsl89:
Link to the notebook:
https://github.com/aws/amazon-sagemaker-examples/blob/master/training/distributed_training/pytorch/data_parallel/mnist/pytorch_smdataparallel_mnist_demo.ipynb

> notebook issue
Model checkpoint cannot be saved based on the logic of the entry point.


aaronmarkham (commented):
I think you should clear the outputs.

aaronmarkham (commented):
What does this fix?

PR #2770, opened 2021-06-25
SageMaker pipelines example of a callBack step followed by a batch transform

georschi:
example for batch transform pipeline with call back step

*Description of changes:*
This notebook introduces an example of a SM Pipeline that makes use of the recently release CallBackStep. 

*Testing done:*
This notebook can run end-to-end within 10 minutes, subject to setting the right execution permissions

## Merge Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your pull request._

- [ x] I have read the [CONTRIBUTING](https://github.com/aws/amazon-sagemaker-examples/blob/master/CONTRIBUTING.md) doc and adhered to the example notebook best practices
- [x ] I have updated any necessary documentation, including [READMEs](https://github.com/aws/amazon-sagemaker-examples/blob/master/README.md)
- [x ] I have tested my notebook(s) and ensured it runs end-to-end
- [ x] I have linted my notebook(s) and code using `tox -e black-format,black-nb-format`

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.


sagemaker-pipelines/tabular/abalone_batch_transform_with_callback_step/utils.py
```diff
@@ -0,0 +1,96 @@
...
+            ModelPackageGroupName=model_package_group_name,
+            ModelApprovalStatus="Approved",
+            SortBy="CreationTime",
+            MaxResults=100,
+        )
+        approved_packages = response["ModelPackageSummaryList"]
+
+        # Fetch more packages if none returned with continuation token
+        while len(approved_packages) == 0 and "NextToken" in response:
+            logger.debug("Getting more packages for token: {}".format(response["NextToken"]))
```
ancamillo, line 70: how do you switch logging level of the logger?
Instead of:
logger = logging.getLogger(__name__)
you could use something like:
logger = logging.getLogger().setLevel(os.getenv("LOGGING_LEVEL", logging.INFO)

so by default the logging level is INFO and if you want to change it to DEBUG for example you can set a lambda env var.

ancamillo (commented):
Great work and thanks for doing this.

Some comments regarding the notebook (can't add them inline for some reason on github):

- A person capable of reading from the queue and the model registry to the return the results to SageMaker -> delete the extra "the"
- For the Pre-Setup section advice to make the lambda trust relationship in plain text so that it can be copy\pasted, make "IAM Management Console" a link to the console, add a cell to get the full name of the sg execution role to edit and probably add couple more screenshots to make it dummy proof
- To get set-up, head over to the IAM Management Console, locate the AmazonSagemaker-ExecutionRole and add the following two permissions -> add colon or full stop at the end?
- For this we are going to use a pretrained model and create and register model to a new Model Registry -> full stop at the end
- To create the layer, we are installing the latest boto3 version in a local folder and we zip this folder in a file that is then used to publish the layer -> ful stop at the end
- *%%capture is added to suppress the output of the following commands sine that is very long. Feel free to comment out to view all output -> since instead of sine
- This event source mapping is triggering the lambda function on arrival of a new message to the SQS queue -> full stop at the end
- empty cell after step 21
- I would add some comments to the code, especially to the cells in the Build Pipeline section
- do you also get "An error occurred loading the pipeline execution graph" when trying to open the pipeline execution graph? maybe the callbackstep not supported yet by the graph.
- I suggest to add something probably after pipeline execution to give an idea that what we were supposed to build actually worked. Also defnitely some comment in the CallbackStep and CreateModelStep code sections to explain the steps. The fact the execution graph didn't work for me makes it hard to visualize what I did, maybe some diagram can be added instead.

georschi:
Hi @ancamillo, 
thanks a lot for your comments, these were some great suggestions!
I've integrated fixes/changes as suggested to those in the commit I just submitted. Let me know if you thing I should be changing something else. 

Specifically for your point: 
> do you also get "An error occurred loading the pipeline execution graph" when trying to open the pipeline execution graph? maybe the callbackstep not supported yet by the graph.

You would need to update Studio to the latest version to be able to see the graph. Thanks to your observation I've added it as an explicit comment so readers of this can know what to do if they get the same error.

Issue #2356, opened 2021-05-13
bert_attention_head_view.ipynb failed CI


hsl89:
Link to the notebook:
https://github.com/aws/amazon-sagemaker-examples/blob/master/sagemaker-debugger/model_specific_realtime_analysis/bert_attention_head_view/bert_attention_head_view.ipynb

Error:

---------------------------------------------------------------------------
Exception encountered at "In [9]":
---------------------------------------------------------------------------
ModuleNotFoundError                       Traceback (most recent call last)
<ipython-input-9-d42a258e5098> in <module>
----> 1 from utils import attention_head_view, neuron_view
      2 from ipywidgets import interactive

/opt/ml/processing/input/utils/attention_head_view.py in <module>
----> 1 from bokeh.plotting import show, figure
      2 from bokeh.models.annotations import Title
      3 from bokeh.models import  ColumnDataSource, Label, Range1d
      4 from bokeh.io import  show, output_notebook, push_notebook
      5 from bokeh.models.glyphs import Line

ModuleNotFoundError: No module named 'bokeh'

PR #3097, opened 2022-01-18
Solve TensorUnavailableStep error for bert_attention_head_view.ipynb

atqy:
*Issue #, if available:*
Fixes https://github.com/aws/amazon-sagemaker-examples/issues/2356
Also, daily CI fails.

*Description of changes:*
Daily CI throws the following error:
```
TensorUnavailableForStep: Value for tensor bertencoder0_transformer0_multiheadattentioncell0_output_1 is not available for step 0 with mode EVAL as it was not saved.
```
It looks like it is trying to access output model tensors when the model hasn't been trained yet. So I made an adjustment to make training finish first before moving on to subsequent cells.

I also fixed two other issues:
* Missing packages (bokeh and ipywidgets)
* Notebook would crash as it stores all the tensors from every step, creating memory problems. Solved this issue by reducing the number of epochs from 3 to 2 as it takes up less space to store data from 2 epochs instead of 3.

*Testing done:* Ran notebook end-to-end in both studio and notebook instance and it works.

## Merge Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your pull request._

- [x] I have read the [CONTRIBUTING](https://github.com/aws/amazon-sagemaker-examples/blob/master/CONTRIBUTING.md) doc and adhered to the example notebook best practices
- [x] I have updated any necessary documentation, including [READMEs](https://github.com/aws/amazon-sagemaker-examples/blob/master/README.md)
- [x] I have tested my notebook(s) and ensured it runs end-to-end
- [x] I have linted my notebook(s) and code using `tox -e black-format,black-nb-format`

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.


sagemaker-debugger/model_specific_realtime_analysis/bert_attention_head_view/bert_attention_head_view.ipynb
```diff
@@ -20,6 +20,16 @@
     "[1] *Visualizing Attention in Transformer-Based Language Representation Models*:  Jesse Vig, 2019, 1904.02679, arXiv"
```
jkroll-aws, line 20: <pre class="ql-syntax" spellcheck="false">Line #4.        while description["SecondaryStatus"] != "Completed":
</pre><p>The above markdown says:</p><blockquote>We can access the tensors from S3 once the training job is in status Training or Completed.</blockquote><p>Your change will fix the out-of-index error, but will take a bit longer because the job must complete first, instead of reaching Training. You could consider/test fixing the list syntax by changing braces to brackets, such as:</p><pre class="ql-syntax" spellcheck="false">while description["SecondaryStatus"] not in ["Training", "Completed"]:
</pre>
<br/>

 _Reply via <a href='https://app.reviewnb.com/aws/amazon-sagemaker-examples/pull/3097/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='13'/>

atqy: Good point, I tried that. The problem is that at the 'Training' stage, there still may not be any data (i.e. training at the first epoch) so the CI error still occurs. After the first epoch is done, like the text says, there are no problems at all. I don't know if there is any way to track when the first epoch is done? What if we keep querying and try...except.. the error until it is gone?

Another concern I have about model not being fully trained is subsequent steps where all the tensors from every epoch is collected. I feel like the model still needs to be fully trained for those steps. Let me know if I am wrong.

jkroll-aws (approved):

jeniyat (approved):
LGTM

